PR #1, opened 2022-06-02
Add MOC ssh keys for core user

larsks:
Add machineconfig resources to configure ssh authorized_keys access for MOC
team.

### DIFF
--- a/cluster-scope/overlays/common/kustomization.yaml
+++ b/cluster-scope/overlays/common/kustomization.yaml
@@ -1,0 +1,5 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+- machineconfigs/99-master-ssh.yaml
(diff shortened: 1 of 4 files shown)


larsks:
@chrisstafford,

This adds ssh keys for the MOC team. Once its merged, you can submit a new PR adding your own keys.

This PR includes two machineconfigs, because I'm not sure if it's possible to have a single machineconfig apply to multiple roles. I know this doesn't matter on the infra cluster, but we'll want to use the same config on the prod cluster.

chrisstafford (approved):
Looks good to me

chrisstafford (approved):

larsks:
@chrisstafford thanks for the review!

In the interests of time, I'm going to be self-merging some changes over the next few days in order to move things forward. On monday's call we can figure out a better process.

PR #5, opened 2022-06-06
feature/argocd rbac

larsks:
This includes three commits:

- Add argocd-allow-all cluster role and binding
- Move all openshift-gitops related config into a bundle
- Enable openshift-gitopts bundle on nerc-ocp-infra

The real purpose is:

Add argocd-allow-all cluster role and binding

This is an attempt to work around [GITOPS-1777][] (aka
redhat-developer/gitops-operator#255). Instead of modifying the rbac
installed by the cluster operator, we're adding our own to avoid triggering
a reconcile.

[GITOPS-1777]: https://issues.redhat.com/browse/GITOPS-1777

### DIFF
--- a/cluster-scope/overlays/nerc-ocp-infra/kustomization.yaml
+++ b/cluster-scope/overlays/nerc-ocp-infra/kustomization.yaml
@@ -2,6 +2,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 resources:
 - ../common
-- ../../base/operators.coreos.com/subscriptions/openshift-gitops-operator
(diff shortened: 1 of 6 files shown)


cluster-scope/overlays/nerc-ocp-infra/kustomization.yaml
```diff
@@ -3,5 +3,5 @@ kind: Kustomization
 resources:
 - ../common
 - ../../base/operators.coreos.com/subscriptions/openshift-gitops-operator
```
naved001: the subscription is part of the bundle

larsks: Argh, that was a bad rebase :). This is why I like reviews :smile:

naved001 (approved):

PR #6, opened 2022-06-06
Feature/cert manager

larsks:
Subscribe to the cert-manager operator on nerc-ocp-infra


naved001 (approved):

mikthoma (approved):
Interesting process. Looks good.

PR #12, opened 2022-06-15
Fix namespace for ACM multiclusterhub

larsks:


mikthoma (approved):
Changed to correct namespace.

PR #20, opened 2022-06-24
Configure bond0 interface

larsks:
We want to use bonded interface pairs on these system. The nodes
aren't yet wired for it, but setting this up now will allow us to
refer to the `bond0` interface in e.g. VLAN configurations (and means
we won't have to re-work those later).

Because we're using OVNKubernetes, we can't use nmstate [1] to enact
the configuration. The recommendation is to apply the configuration
using a MachineConfig [2] resource, but this is complicated by the
fact that our nodes don't all have the same interface names, and it's
not possible to apply node-specific machineconfigs [3].

We work around this solution by:

1. Copying nmconnection files for *all hosts* to *every host*, but
  placing them in `/etc/mco` (just because that's a convenient
  available directory, it seems relatively topical, and it's not
  possible to create new directories using the `directories` section
  of an ignition config).

2. Installing a systemd unit that runs a shell script at boot that
  copies the host-specific configs from `/etc/mco` into
  `/etc/NetworkManager/system-connections`.

[1]: https://docs.openshift.com/container-platform/4.10/networking/k8s_nmstate/k8s-nmstate-about-the-k8s-nmstate-operator.html
[2]: https://docs.openshift.com/container-platform/4.10/post_installation_configuration/machine-configuration-tasks.html
[3]: https://github.com/openshift/machine-config-operator/issues/1720


larsks:
Because we're working around limitations in the MachineConfig operator, this is a pretty hairy solution. While I've been trying other things out "live", I really want some eyes on this before I push it out.

cluster-scope/overlays/nerc-ocp-infra/machineconfigs/configure-bond0/Makefile
```diff
@@ -0,0 +1,10 @@
+MACHINECONFIGS = \
```
dystewart, line 1: Where is this makefile being used? Maybe I'm missing where it runs

cluster-scope/overlays/nerc-ocp-infra/machineconfigs/configure-bond0/src/bond0.nmconnection
```diff
@@ -0,0 +1,17 @@
+[connection]
```
dystewart, line 1: Connection profile of the main bond0 interface

cluster-scope/overlays/nerc-ocp-infra/machineconfigs/configure-bond0/src/bond0.nmconnection
```diff
@@ -0,0 +1,17 @@
...
+type=bond
+interface-name=bond0
+autoconnect=true
+connection.autoconnect-slaves=1
+
+[ethernet]
+mtu=9000
+
+[bond]
+mode=balance-rr
```
dystewart, line 12: Do we need to use this mode since the profiles are all using bond0 as the master?

cluster-scope/overlays/nerc-ocp-infra/machineconfigs/configure-bond0/src/bond0.nmconnection
```diff
@@ -0,0 +1,17 @@
...
+miimon=140
```
dystewart, line 13: what is this pair specifying?

cluster-scope/overlays/nerc-ocp-infra/machineconfigs/configure-bond0/src/configure-bond0-ports.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+mkdir -p /etc/NetworkManager/system-connections-disabled
+mv /etc/NetworkManager/system-connections/ens*.nmconnection /etc/NetworkManager/system-connections-disabled/
```
dystewart, line 4: Disable default or existing connection profiles on the host?

cluster-scope/overlays/nerc-ocp-infra/machineconfigs/configure-bond0/src/configure-bond0.bu
```diff
@@ -0,0 +1,50 @@
+variant: openshift
```
dystewart, line 1: This is the file that butane will use to auto-generate the machineConfig file. So when the machineConfig is generated and applied the connection profiles and the systemd service below will be

cluster-scope/overlays/nerc-ocp-infra/machineconfigs/configure-bond0/src/configure-bond0.bu
```diff
@@ -0,0 +1,50 @@
...
+      contents: |
+        [Unit]
+        Description = Apply node-specific network configuration for bond0
+        RequiredBy = NetworkManager.service
+        After = network-pre.target
+        Before = NetworkManager.service
+
+        [Service]
+        Type = oneshot
+        ExecStart = /etc/mco/configure-bond0-ports.sh
```
dystewart, line 47: Does this mean the script is run on system start? Or when the service itself starts?

cluster-scope/overlays/nerc-ocp-infra/machineconfigs/configure-bond0/src/ctl-0-nic1.nmconnection
```diff
@@ -0,0 +1,10 @@
+[connection]
+id=ens6f0
+type=ethernet
+interface-name=ens6f0
+master=bond0
```
dystewart, line 5: In the ethernet section there is no MAC address specified, as there is in the examples at: [https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/configuring_and_managing_networking/assembly_manually-creating-networkmanager-profiles-in-key-file-format_configuring-and-managing-networking](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/configuring_and_managing_networking/assembly_manually-creating-networkmanager-profiles-in-key-file-format_configuring-and-managing-networking). Is this omission possible because we are specifying the bond0 profile here which is created above?

dystewart (commented):
Going to leave some questions/statements throughout to make sure I'm understanding the networking (which I'm not very familiar with)

cluster-scope/overlays/nerc-ocp-infra/machineconfigs/configure-bond0/Makefile
larsks: You run it locally if you make changes to anything in the `src/` directory.

cluster-scope/overlays/nerc-ocp-infra/machineconfigs/configure-bond0/src/bond0.nmconnection
larsks: Was this a question, or a suggestion for documentation, or....?

cluster-scope/overlays/nerc-ocp-infra/machineconfigs/configure-bond0/src/bond0.nmconnection
larsks: The `mode` parameter controls the configuration of the bond itself (whether using Etherchannel (which is `balance-rr`) or LACP (`mode=802.3ad`). See e.g. https://www.nodeum.io/howto/lacp-bonding-configuration-by-using-nmcli-centos-or-rhel-7.

cluster-scope/overlays/nerc-ocp-infra/machineconfigs/configure-bond0/src/configure-bond0-ports.sh
larsks: Right; this removes the existing interface configuration for the `ens6f0` or `ens5f0` interface.

cluster-scope/overlays/nerc-ocp-infra/machineconfigs/configure-bond0/src/configure-bond0.bu
larsks: A system unit, when enabled, runs at system start. The dependencies in the `[Unit]` section mean that we expect this to run *before* NetworkManager starts up.

cluster-scope/overlays/nerc-ocp-infra/machineconfigs/configure-bond0/src/ctl-0-nic1.nmconnection
larsks: I think in those examples the MAC address is being used to identify the interface. In this case, we're using the `interface-name` parameter. It wouldn't make sense to specify a MAC address here because this profile is expected to apply on multiple machines.

cluster-scope/overlays/nerc-ocp-infra/machineconfigs/configure-bond0/src/configure-bond0.bu
larsks: This comment seems truncated, but it's correct so far: This is the Butane input file; the `Makefile` uses the `butane` command to produce the `MachineConfig` resource, which is that we actually reference in the `kustomization.yaml` file.

cluster-scope/overlays/nerc-ocp-infra/machineconfigs/configure-bond0/src/bond0.nmconnection
larsks: I've already covered the `mode` parameter; the `miimon` parameter specifies the link monitoring interval. From [these docs](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/networking_guide/sec-using_channel_bonding):

> `miimon=time_in_milliseconds`
>
> Specifies (in milliseconds) how often MII link monitoring occurs. This is useful if high availability is required because MII is used to verify that the NIC is active. To verify that the driver for a particular NIC supports the MII tool, type the following command as root:

I'm using this particular value because that's what @jtriley used when doing the test install on these systems.

larsks:
@dystewart thanks for the questions; hope my answers are clear. Let me know if not!

larsks:
An alternative solution here might be to be [disable "predictable device names"](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/networking_guide/sec-disabling_consistent_network_device_naming), because they don't appear to be predictable. In *theory*, the device name is based on the PCI bus/slot/etc, but in this case all three control nodes have an identical PCI configuration:

```
# lspci | grep -i ethernet
06:00.0 Ethernet controller: Mellanox Technologies MT27710 Family [ConnectX-4 Lx]
06:00.1 Ethernet controller: Mellanox Technologies MT27710 Family [ConnectX-4 Lx]
31:00.0 Ethernet controller: Intel Corporation Ethernet Connection X722 for 10GBASE-T (rev 09)
31:00.1 Ethernet controller: Intel Corporation Ethernet Connection X722 for 10GBASE-T (rev 09)
```

But for whatever reason, the interface base name is `ens6` on ctl-0 and `ens5` on ctl-1 and ctl-2.

larsks:
I think that [disabling predicable interface names](https://github.com/OCP-on-NERC/nerc-ocp-config/pull/20#issuecomment-1166060119) is the right idea. I've submitted an alternative implementation of this PR in #22. This is a dramatically simpler change, and some limited testing suggests that it works as advertised.

larsks:
I've removed all the reviews and converted this PR to a draft because I think we'll end up going with #22.

PR #22, opened 2022-06-25
Configure bond0 interface (alternative implementation)

larsks:
This is an alternative implementation of #20.

We want to use bonded interface pairs on these system. The nodes
aren't yet wired for it, but setting this up now will allow us to
refer to the `bond0` interface in e.g. VLAN configurations (and means
we won't have to re-work those later).

Because we're using OVNKubernetes, we can't use nmstate [1] to enact
the configuration. The recommended mechanism for modifying the primary
interface configuration is to apply the configuration using a
MachineConfig [2] resource.

Despite having identical PCI configurations, ctl-0 has different
network device names than ctl-1 and ctl-2. By disabling "predictable
network interface names" [3], we end up with consistent device names
across the cluster.

By disabling predictable network interface names, we can
simply install NetworkManager connection files for bond0, eth2, and
eth3. This will take effect when the system reboots (which will happen
when the MachineConfig Operator applies our changes).

[1]: https://docs.openshift.com/container-platform/4.10/networking/k8s_nmstate/k8s-nmstate-about-the-k8s-nmstate-operator.html
[2]: https://docs.openshift.com/container-platform/4.10/post_installation_configuration/machine-configuration-tasks.html
[3]: https://www.freedesktop.org/wiki/Software/systemd/PredictableNetworkInterfaceNames/


cluster-scope/overlays/nerc-ocp-infra/machineconfigs/configure-bond0/machineconfig.yaml
```diff
@@ -0,0 +1,26 @@
+# Generated by Butane; do not edit
+apiVersion: machineconfiguration.openshift.io/v1
+kind: MachineConfig
+metadata:
+  labels:
+    machineconfiguration.openshift.io/role: master
```
naved001, line 6: just wanted to double check that we are only doing this for the controllers

larsks: This is for the infra cluster, which at the moment only has controllers.

naved001 (approved):

cluster-scope/overlays/nerc-ocp-infra/machineconfigs/configure-bond0/src/bond0.nmconnection
```diff
@@ -0,0 +1,17 @@
...
+type=bond
+interface-name=bond0
+autoconnect=true
+connection.autoconnect-slaves=1
+
+[ethernet]
+mtu=9000
+
+[bond]
+mode=balance-rr
```
jtriley: We typically use mode 802.3ad for LACP

jtriley (changes requested):
LGTM but let's switch the LACP bond mode to 802.3ad

cluster-scope/overlays/nerc-ocp-infra/machineconfigs/configure-bond0/src/bond0.nmconnection
larsks: @jtriley  I was going to ask about that; the [example you showed me last week](https://github.com/nerc-project/nerc-k8s-operators/blob/nerc-ocp-prod/k8s/overlays/nerc-ocp-infra/net-configs/ctl-0-bond.yaml) was using balance-rr, which is why I used it here.

jtriley: Yeah that crept in from copy/paste of nmstate bond config example. I later changed that in test deployment of nerc-ocp-test.

jtriley (approved):

PR #26, opened 2022-06-28
Add Hashicorp vault manifests to nerc-ocp-infra cluster

dystewart:
This PR adds the necessary manifests for Hashicorp Vault installation. The ArgoCD app will be added to the nerc-ocp-apps repo

### DIFF
--- a/cluster-scope/base/core/namespaces/vault/kustomization.yaml
+++ b/cluster-scope/base/core/namespaces/vault/kustomization.yaml
@@ -1,0 +1,4 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+- namespace.yaml
--- a/vault/README.md
+++ b/vault/README.md
@@ -1,0 +1,8 @@
+# Hashicorp Vault
+This Vault deployment was derived from the official Hashicorp Vault Helm release at: [Hashicorp](https://github.com/hashicorp/vault-helm)
+
+There are some differences to the out of the box release however, for instance the vault agent injector is not utilized in the OCP-on-NERC configuration as we are using External Secrets to fetch secrets from the Vault.
+
+The base of this Vault deployment was produced after adding the helm repository with: `helm repo add hashicorp https://helm.releases.hashicorp.com`
+
+Then the underlying yalm manifests were generated with `helm template vault hashicorp/vault`
--- a/vault/base/configmaps/configmap.yaml
+++ b/vault/base/configmaps/configmap.yaml
@@ -1,0 +1,23 @@
+apiVersion: v1
+data:
+  extraconfig-from-values.hcl: |-
+    disable_mlock = true
+    ui = true
+    listener "tcp" {
+      tls_disable = 1
+      address = "[::]:8200"
+      cluster_address = "[::]:8201"
+    }
+    storage "raft" {
+      path = "/vault/data"
+    }
+    service_registration "kubernetes" {}
+kind: ConfigMap
+metadata:
+  labels:
+    app.kubernetes.io/instance: nerc
+    app.kubernetes.io/managed-by: Helm
+    app.kubernetes.io/name: vault
+    helm.sh/chart: vault-0.18.0
+  name: vault-config
+  namespace: vault
--- a/vault/base/statefulsets/statefulset.yaml
+++ b/vault/base/statefulsets/statefulset.yaml
@@ -1,0 +1,145 @@
+apiVersion: apps/v1
+kind: StatefulSet
+metadata:
+  labels:
+    app.kubernetes.io/instance: nerc
+    app.kubernetes.io/managed-by: Helm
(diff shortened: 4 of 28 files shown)


larsks:
@dystewart you have a `kustomization` instead of `kustomization.yaml` in vault/base/services.

vault/base/statefulsets/statefulset.yaml
```diff
@@ -0,0 +1,145 @@
...
+      serviceAccountName: vault
+      volumes:
+        - name: config
+          configMap:
+            name: vault-config
+        - name: home
+          emptyDir: {}
+      containers:
+        - name: vault
+          image: quay.io/operate-first/vault:1.9.2
```
dystewart: Is there a registry we can use to store this image related to NERC?

cluster-scope/base/core/namespaces/vault/kustomization.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+- namespace.yaml
```
dystewart, line 4: @larsks in OPF we give the namespace resource quotas etc here. How are those decided in this environment?

larsks: We haven't gotten that far yet :). Don't worry about it for now; we'll add in quotas etc. in a future PR.

vault/base/statefulsets/statefulset.yaml
larsks: @dystewart Tell me more; what is this image for? Why do we need a custom image rather than the upstream `hashicorp/vault:1.10.3` image?

larsks (commented):
When we're importing third-party manifests like this, I think it would be a good idea to document where it all came from: can you add a README to the `vault` directory that describes how  you got from whatever Hashicorp provides to the manifests included in this PR?

vault/base/statefulsets/statefulset.yaml
dystewart: @larsks I'm asking because we had to pull the image from Docker and upload it to quay because of docker pull rate issues, which were preventing the image from being pulled successfully on deployment in OperateFirst. With Quay we don't have this rate limiting.  
It's a bit behind the current version bc we have yet to update it in OPF. I'll look into what has been added/changed in the newer versions though. It may make more sense to use the newer version, given I'm able to quickly get it up and confirm things work as currently constituted

larsks: I think I would just point this at the upstream image. If pull rate limits become an issue, we can explore solutions at that point (for example, maybe we would manually push the image into the openshift container registry...).

dystewart:
@larsks updated the image to upstream and added README.md file to vault dir

naved001 (approved):
looks good to me.

vault/base/configmaps/configmap.yaml
```diff
@@ -0,0 +1,23 @@
...
+      cluster_address = "[::]:8201"
+    }
+    storage "raft" {
+      path = "/vault/data"
+    }
+    service_registration "kubernetes" {}
+kind: ConfigMap
+metadata:
+  labels:
+    app.kubernetes.io/instance: opf
```
larsks: Some of these resources have `app.kubernetes.io/instance: opf` and others have `app.kubernetes.io/instance: nerc`. We should probably just remove these from the individual manifests and apply them via the top-level `kustomization.yaml`.

If not that, we should at least ensure that they are all consistent.

vault/base/services/kustomization
```diff
@@ -0,0 +1,8 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
```
larsks: I think this file should be removed from the PR.

larsks (changes requested):

vault/README.md
```diff
@@ -0,0 +1,8 @@
+# Hashicorp Vault
+This Vault deployment was derived from the official Hashicorp Vault Helm release at: [Hashicorp](https://github.com/hashicorp/vault-helm)
+
+There are some differences to the out of the box release however, for instance the vault agent injector is not utilized in the OCP-on-NERC configuration as we are using External Secrets to fetch secrets from the Vault. 
+
+The base of this Vault deployment was produced after adding the helm repository with: `helm repo add hashicorp https://helm.releases.hashicorp.com`
+
+Then the undelying yalm manifests were generated with `helm template vault hashicorp/vault`
```
dystewart: fix spelling

dystewart:
@larsks Removed extra kustomize file and changed opf to nerc where applicable. Also squashed commits

naved001:
@dystewart looks like there are some merge conflicts that need fixing.

joachimweyl:
@larsks are the changes that @dystewart made going to resolve the change requests?

larsks (approved):

PR #27, opened 2022-07-01
Configure OpenShift image registry

larsks:
Now that we have storage we can enable the image registry. Because
we're limited to Ceph RBD storage, we only have RWO [1] volumes, which
means we can only have a single replica for the registry service.

It is possible to use an S3 backend for the registry (which would
allow us to have multiple replicas), but it looks as if getting things
set up to use a bucket from ODF [2] will be tricky.

[1]: https://kubernetes.io/docs/concepts/storage/persistent-volumes/#access-modes
[2]: https://www.redhat.com/en/technologies/cloud-computing/openshift-data-foundation

### DIFF
--- a/cluster-scope/base/core/persistentvolumeclaims/image-registry-storage/persistentvolumeclaim.yaml
+++ b/cluster-scope/base/core/persistentvolumeclaims/image-registry-storage/persistentvolumeclaim.yaml
@@ -1,0 +1,11 @@
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: image-registry-storage
+  namespace: openshift-image-registry
+spec:
+  accessModes:
+    - ReadWriteOnce
+  resources:
+    requests:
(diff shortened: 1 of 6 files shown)


computate (approved):
Good to know that the OpenShift Image Registry requires 100Gi.

cluster-scope/base/core/persistentvolumeclaims/image-registry-storage/persistentvolumeclaim.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
```
naved001, line 3: does this claim need to be in namespace `openshift-image-registry`?

larsks: Yes!

PR #33, opened 2022-07-14
Configure ESO instance for use with Vault

dystewart:
This will create an instance of ESO using the default OperatorConfig provided by ESO operator.


larsks:
@dystewart  it turns out you and I were working on the same thing; see #34 which both sets up the operatorconfig and creates a clustersecretstore that authenticates against vault using a kubernetes serviceaccount.

joachimweyl:
@larsks @dystewart is out of the office today. It looks like your version is working. Shall we use that and move forward with reviews or would you prefer to review @dystewart's version and merge that since on yours you stated it was a draft? I also noticed there are some pretty big differences between the two so maybe neither of my suggestions is the correct direction to go.

larsks:
@joachimweyl I'd like to discuss with @dystewart before we move on either one of these.

larsks:
Closed in favor of #34.

PR #34, opened 2022-07-15
Configure ClusterSecretStore for infra cluster

larsks:
A ClusterSecretStore [1] creates a cluster-scoped secret store that will watch for ExternalSecret resources in all namespaces.

This commit creates a ClusterSecretStore that by defaults points at the vault instance on the local cluster and authenticates using a ServiceAccount. Every cluster overlay will need to patch spec.provider.vault.auth.kubernetes.mountPath, and clusters other than nerc-ocp-infra will need to (a) patch spec.provider.vault.server and (b) provide an alternative auth mechanism.

[1]: https://external-secrets.io/latest/api-clustersecretstore/

---
  - https://app.asana.com/0/0/1202594685849610

### DIFF
--- a/cluster-scope/overlays/nerc-ocp-infra/clustersecretstores/nerc-cluster-secrets_patch.yaml
+++ b/cluster-scope/overlays/nerc-ocp-infra/clustersecretstores/nerc-cluster-secrets_patch.yaml
@@ -1,0 +1,10 @@
+apiVersion: external-secrets.io/v1beta1
+kind: ClusterSecretStore
+metadata:
+  name: nerc-cluster-secrets
+spec:
+  provider:
+    vault:
+      auth:
+        kubernetes:
+          mountPath: kubernetes/nerc-ocp-infra/
--- a/cluster-scope/base/core/namespaces/external-secrets-operator/kustomization.yaml
+++ b/cluster-scope/base/core/namespaces/external-secrets-operator/kustomization.yaml
@@ -1,5 +1,4 @@
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
-namespace: external-secrets-operator
 resources:
-- subscription.yaml
+- namespace.yaml
--- a/cluster-scope/base/core/namespaces/external-secrets-operator/namespace.yaml
+++ b/cluster-scope/base/core/namespaces/external-secrets-operator/namespace.yaml
@@ -1,0 +1,5 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: external-secrets-operator
+spec: {}
--- a/cluster-scope/base/core/serviceaccounts/eso-vault-auth/kustomization.yaml
+++ b/cluster-scope/base/core/serviceaccounts/eso-vault-auth/kustomization.yaml
@@ -1,0 +1,5 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+namespace: external-secrets-operator
+resources:
+    - serviceaccount.yaml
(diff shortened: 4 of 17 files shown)


larsks:
Marked as a draft right now because it's late at night and the description needs work, but it seems to be working.

naved001 (approved):

larsks:
Converting this to draft again while reviewing docs at https://github.com/operate-first/apps/tree/c51fcbfc67f997d6b5c28c255a9acb530b56a7ed/docs/content/vault_eso

cluster-scope/base/core/serviceaccounts/vault-nerc-secret-reader/serviceaccount.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: vault-nerc-secret-reader
+  namespace: external-secrets-config
```
dystewart: In OPF we actually don't have a vault-secret-reader sa in the ESO namespace. We do have vault-secret-reader in each namespace which utilizes vault. We will need a token-reviewer in the  external-secrets-config namespace with auth delegtion crb:
```
kind: ClusterRoleBinding
metadata:
  name: eso-tokenreview
roleRef:
  kind: ClusterRole
  name: system:auth-delegator
subjects:
  - kind: ServiceAccount
    name: eso-vault-auth
    namespace: external-secrets-config
```

dystewart: That's not to say we can't do it this way though

larsks: > We will need a token-reviewer in the external-secrets-config namespace with auth delegtion crb:

We don't need this on the cluster in which Vault is installed (at least, it all seems to work just fine without that). We will need both this RBAC and additional configuration for other clusters.

> In OPF we actually don't have a vault-secret-reader sa in the ESO namespace

For the `nerc-ocp-infra` cluster, I think it makes sense to just have a single `ClusterSecretStore` (in which case we just need a namespace to hold the `ServiceAccount`  used by the `ClusterSecretStore`). On the prod cluster we will probably want per-namespace `SecretStores` instead (to prevent us from accidentally exposing secrets to end-user namespaces).

larsks:
I've substantially refactored this pr. In particular:

- It moves the external secrets operator into its own namespace (currently, it's installed into the `openshift-operators` namespace)
- It includes the RBAC we'll need to support serviceaccount-based auth on clusters other than nerc-ocp-infra

cluster-scope/base/core/serviceaccounts/vault-nerc-secret-reader/serviceaccount.yaml
dystewart: Yeah agreed I think this makes sense. Also simplifies infra a bit and only admins should have access to this cluster anyway if I remember correctly. 
I also think in the prod cluster per namespace is the way to go.

cluster-scope/overlays/nerc-ocp-infra/clustersecretstores/nerc-cluster-secrets_patch.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: external-secrets.io/v1beta1
+kind: ClusterSecretStore
+metadata:
+  name: nerc-cluster-secrets
+spec:
+  provider:
```
dystewart, line 6: @larsks Oops I guess I didn't submit this comment yesterday. Is this configuration set up internally in Vault? If not I would be happy to get that started

larsks:
> Is this configuration set up internally in Vault? If not I would be happy to get that started

It's partially in place in Vault. I'm trying to make sure the vault configuration is automated, which I've been working on [here](https://github.com/OCP-on-NERC/vault-config).

larsks:
@dystewart I think this is ready to merge. Let me know if you see anything else that needs changing?

dystewart (approved):

dystewart:
@larsks this looks good let me know if you need any assistance with vault config

PR #44, opened 2022-07-27
adds nerc-ocp-prod overlay

chrisstafford:
---
  - https://app.asana.com/0/0/1202594881550879


computate:
Looks like the kustomization.yaml file in this PR is empty, would that work correctly?

naved001:
yeah the kustomization file should refer to `clusterversion.yaml`

chrisstafford:
Must not have commited that file properly. It should be good now

computate (approved):

PR #45, opened 2022-07-28
adds api and app certs for prod and api for infra

chrisstafford:
---
  - https://app.asana.com/0/0/1202594685849630
  - https://app.asana.com/0/0/1201794021034508

### DIFF
--- a/cluster-scope/overlays/nerc-ocp-infra/apiserver/default_patch.yaml
+++ b/cluster-scope/overlays/nerc-ocp-infra/apiserver/default_patch.yaml
@@ -1,0 +1,8 @@
+apiVersion: operator.openshift.io/v1
+kind: APIServer
+metadata:
+  name: cluster
+  namespace: openshift-config
+spec:
+  defaultCertificate:
+    name: default-api-certificate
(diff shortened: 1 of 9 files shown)


chrisstafford:
Clusterversion is included in this PR: https://github.com/OCP-on-NERC/nerc-ocp-config/pull/44

naved001:
I am a bit worried about adding things for the nerc-ocp-prod cluster because it's not up and running. Generally, I like to go into argocd after a PR is merged to see if the changes were applied correctly (sometimes we end up making typos, so we correct those). If we accumulate all the manifests to apply to the nerc-ocp-prod cluster I suspect we'll be spending lots of time in argocd figuring out why it's not syncing properly. Just my thoughts, but if people feel reasonably confident then we should go ahead.

Also I think it might be better if this PR was split into different commits (if not multiple PRs). A single commit that creates certificates for multiple environments and clustersecretstore for one might be bit much.

chrisstafford:
Fully understandable. I can split this out into multiple PRs and resubmit, we can also hold off merge on the prod cluster until it's up and running.

PR #38, opened 2022-07-22
Install infra ingress certificate

larsks:
Install ssl certificate for `*.apps.nerc-ocp-infra.rc.fas.harvard.edu`.

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1201794020676126

PR #46, opened 2022-07-29
Adds api certificate to infra overlay

chrisstafford:
---
  - https://app.asana.com/0/0/1201794021034508


jtriley (changes requested):
This is failing CI/CD for same reasons as #47

cluster-scope/base/config.openshift.io/apiserver/cluster/apiserver.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: config.openshift.io/v1
+kind: APIServer
+metadata:
+  annotations:
+    include.release.openshift.io/ibm-cloud-managed: "true"
+    include.release.openshift.io/self-managed-high-availability: "true"
+    include.release.openshift.io/single-node-developer: "true"
+    oauth-apiserver.openshift.io/secure-token-storage: "true"
+    release.openshift.io/create-only: "true"
+  name: cluster
```
dystewart: @chrisstafford I think the issue is this resource in the infra cluster: [https://console-openshift-console.apps.nerc-ocp-infra.rc.fas.harvard.edu/k8s/cluster/config.openshift.io~v1~APIServer/cluster/yaml](https://console-openshift-console.apps.nerc-ocp-infra.rc.fas.harvard.edu/k8s/cluster/config.openshift.io~v1~APIServer/cluster/yaml)
It has the same name as the `APIserver` you're adding here. It's in the cluster but it's not in the repo(as far as I can tell) so maybe it was added via the console already. 

I think by removing just this file and just patching the existing resource will fix the issue but maybe the original resource should be deleted to add this `APIserver` resource

chrisstafford: Hmmm, I believe the APIServer is created on install, looking at the operatefirst cluster repo I only see a patch file for the APIServer and nothing in the base. I think the move here would be to remove the base file and edit the existing resource.

joachimweyl: @chrisstafford did this resolve the issue?

dystewart (approved):

joachimweyl:
> This is failing CI/CD for same reasons as #47

@chrisstafford is this change request resolved?

chrisstafford:
> This is failing CI/CD for same reasons as #47

@jtriley, I was able to resolve the issue another way, could you pull back your requested changes so we can merge?

cluster-scope/overlays/nerc-ocp-infra/kustomization.yaml
```diff
@@ -22,6 +22,7 @@ resources:
 - clusterrolebindings/nerc-ops-sudoers.yaml
 - clusterrolebindings/nerc-ops-portforward.yaml
 - externalsecrets
+- apiserver/api_server_cert.yaml
```
larsks: It's important to note that this isn't just setting the API certificate: it's unilaterally defining what the cluster APIServer resource looks like. This is why in #38 we create a `IngressController` resource in `base`, and then *patch* in the cluster overlay (so that we're clearly separating "this is what the resource should look like" from "this file just adds a certificate").

I'm not going to insist we do it the same way here, but we should rename this file to simply use the name of the resource it is creating, so `apiserver/cluster.yaml`.

larsks (changes requested):

cluster-scope/overlays/nerc-ocp-infra/kustomization.yaml
chrisstafford: I appreciate the insight @larsks I've updated the filename as requested.

joachimweyl:
@chrisstafford what are the next steps?

chrisstafford:
> @chrisstafford what are the next steps?

It looks like github still needs @jtriley's requested changes to be marked as resolved.

joachimweyl:
@jtriley is this something you can review today? This is one of the last Infra install steps.

joachimweyl:
@larsks I see that your review is not complete. It appears that @chrisstafford made changes that outdated your request. Does @chrisstafford need to do anything else to get to an approved PR?

larsks (approved):

jtriley (approved):

PR #50, opened 2022-08-05
External ingress

chrisstafford:
---
  - https://app.asana.com/0/0/1202594685849644


chrisstafford:
I still need to create the secrets for the certificate and still need clarity on whether or not we want an API endpoint exposed outside the network. I would however like a sanity check that this at least looks correct.

cluster-scope/overlays/nerc-ocp-prod/clustersecretstores/nerc-cluster-secrets_patch.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: external-secrets.io/v1beta1
```
larsks, line 1: For the production cluster,  we're going to want per-namespace secretstores. A cluster-wide secret store would potentially expose sensitive secrets to all cluster users.

cluster-scope/overlays/nerc-ocp-prod/ingresscontrollers/external-apps.ingress-controller.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: operator.openshift.io/v1
```
larsks, line 1: This ingress controller needs to be configured to listen on a public address. That means we're first going to need to configure an appropriate interface on all the worker nodes (so probably a `NodeNetworkConfigurationPolicy` resource), and then we're going to need to perform some work -- of which I don't know the specifics -- to bind the ingress controller to the public address.

larsks (changes requested):
I've requested one specific change, but I'd like to make a couple more broad suggestions.

I prefer to see pull requests on a particular feature (e.g., this one would be "Configure external ingress"). Things that are dependencies of that feature, but not directly related (like configuring secret stores, or configuring the API certificate) should be split out into separate pull requests. This has a bunch of advantages:

- Smaller pull requests are easier to review
- It allows us to merge the dependencies while still working on other parts of the PR
- It may enable other folks to work on unrelated pull requests that share the same dependencies

chrisstafford:
@larsks, I agree on all points, I just branched this from a branch still waiting to be merged. I'll close this PR and create a new branch from the main and include your changes.

naved001:
@chrisstafford should this be closed?

joachimweyl:
@chrisstafford ^

PR #52, opened 2022-08-10
Increase database storage size for ACM agent service

larsks:
The database PV size was set to 100M, which ends up being too small. This
commit updates the requested size in the AgentServiceConfig. It's not clear
if this will also update the existing PVC or if we'll need to do that
separately.


joachimweyl (approved):
Approved. Do we have an idea of how fast this will grow?

larsks:
@joachimweyl if you mean "how long will the resize take?", the answer is "instantaneously". If you mean, "at what rate is new data added to this volume?", we don't know.

joachimweyl:
@larsks I was referring to at what rate new data is added. I guess we will know next time we need to expand it.

PR #53, opened 2022-08-11
Forcibly refresh the storage network interface every three days

larsks:
Add a systemd timer [1] and service [2] that will forcibly refresh the
bond0.2177 interface every three days (by bringing it down and then back up
using nmcli).

This is a workaround for rhbz#2105088 [3].

[1]: https://www.freedesktop.org/software/systemd/man/systemd.timer.html
[2]: https://www.freedesktop.org/software/systemd/man/systemd.service.html
[3]: https://bugzilla.redhat.com/show_bug.cgi?id=2105088

### DIFF
--- a/cluster-scope/overlays/nerc-ocp-infra/machineconfigs/refresh-storage-interface.yaml
+++ b/cluster-scope/overlays/nerc-ocp-infra/machineconfigs/refresh-storage-interface.yaml
@@ -1,0 +1,34 @@
+apiVersion: machineconfiguration.openshift.io/v1
+kind: MachineConfig
+metadata:
+  labels:
+    machineconfiguration.openshift.io/role: master
+  name: refresh-storage-interface
+spec:
+  config:
+    ignition:
+      version: 3.2.0
+  systemd:
+    units:
+      - name: refresh-storage-interface.timer
+        enabled: true
+        contents: |
+          [Timer]
+          OnBootSec=3d
+          OnUnitActiveSec=3d
+          RandomizedDelaySec=4h
+
+          [Install]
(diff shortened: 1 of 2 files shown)


larsks:
@joachimweyl assuming this merges over the next week you won't need to remind us about this issue next Friday.

cluster-scope/overlays/nerc-ocp-infra/machineconfigs/refresh-storage-interface.yaml
```diff
@@ -0,0 +1,34 @@
...
+        enabled: true
+        contents: |
+          [Timer]
+          OnBootSec=3d
+          OnUnitActiveSec=3d
+          RandomizedDelaySec=4h
+
+          [Install]
+          WantedBy=timers.target
+      - name: refresh-storage-interface.service
```
naved001, line 23: I guess the service doesn't need to be enabled when there's a corresponding timer for it?

naved001 (approved):
looks good, just one question since I am not familiar with systemd timers

cluster-scope/overlays/nerc-ocp-infra/machineconfigs/refresh-storage-interface.yaml
knikolla: Correct. According to documentation:
> For each timer file, a matching unit file must exist, describing the unit to activate when the timer elapses. By default, a service by the same name as the timer (except for the suffix) is activated.

joachimweyl:
removed auto reminders from Slack.

PR #57, opened 2022-08-17
Fix namespace in ExternalSecret for external ceph cluster

naved001:
The ExternalSecret was not created due to a typo in the word "namespace".
The actual secret exists which may have been created manually.

Part of trying to fix argocd sync issues.


naved001:
closing this in favor of #58

PR #61, opened 2022-08-18
increases fs and img storage for acm installs

chrisstafford:


jtriley (approved):

mikthoma:
LGTM, I think the 4.10 and 4.11 images are exactly what we want in there for now.

mikthoma (commented):
Look good!

PR #63, opened 2022-08-22
Integrate with NERC Keycloak

knikolla:
Client definition created in https://github.com/nerc-project/mss-keycloak/blob/main/k8s/overlays/nerc-shift-1/clients/ocp-prod.yaml

---
  - https://app.asana.com/0/0/1202099997018465

### DIFF
--- a/cluster-scope/base/config.openshift.io/oauths/cluster/oauth.yaml
+++ b/cluster-scope/base/config.openshift.io/oauths/cluster/oauth.yaml
@@ -7,4 +7,5 @@ metadata:
     include.release.openshift.io/single-node-developer: "true"
     release.openshift.io/create-only: "true"
   name: cluster
-spec: {}
+spec:
+  identityProviders: []
--- a/cluster-scope/overlays/nerc-ocp-prod/externalsecrets/oauths-clientsecret-nerc.yaml
+++ b/cluster-scope/overlays/nerc-ocp-prod/externalsecrets/oauths-clientsecret-nerc.yaml
@@ -1,0 +1,16 @@
+apiVersion: external-secrets.io/v1beta1
+kind: ExternalSecret
+metadata:
+  name: oauths-clientsecret-nerc
+  namespace: openshift-config
+spec:
+  secretStoreRef:
+    kind: SecretStore
+    name: nerc-secret-store
+  target:
+    name: oauths-clientsecret-nerc
+  data:
+    - secretKey: secret
+      remoteRef:
+        key: nerc/nerc-ocp-prod/openshift-config/oauths-clientsecret-nerc
+        property: secret
--- a/cluster-scope/overlays/nerc-ocp-prod/kustomization.yaml
+++ b/cluster-scope/overlays/nerc-ocp-prod/kustomization.yaml
@@ -14,6 +14,7 @@ resources:
 
 patches:
 - path: ingresscontrollers/default_patch.yaml
+- path: oauths/cluster_patch.yaml
 
 - target:
     group: "external-secrets.io"
--- a/cluster-scope/overlays/nerc-ocp-prod/oauths/cluster_patch.yaml
+++ b/cluster-scope/overlays/nerc-ocp-prod/oauths/cluster_patch.yaml
@@ -1,0 +1,22 @@
+apiVersion: config.openshift.io/v1
+kind: OAuth
+metadata:
+  name: cluster
+spec:
+  identityProviders:
+    - mappingMethod: claim
+      name: mss-keycloak
+      openID:
+        claims:
+          email:
+            - email
+          name:
+            - name
+          preferredUsername:
+            - preferred_username
+        clientID: ocp-prod
+        clientSecret:
+          name: oauths-clientsecret-nerc
+        extraScopes: []
+        issuer: https://keycloak.mss.mghpcc.org/auth/realms/mss
+      type: OpenID
--- a/cluster-scope/overlays/nerc-ocp-prod/externalsecrets/kustomization.yaml
+++ b/cluster-scope/overlays/nerc-ocp-prod/externalsecrets/kustomization.yaml
@@ -3,3 +3,4 @@ kind: Kustomization
 resources:
 - default-ingress-certificate.yaml
 - default-api-certificate.yaml
+- oauths-clientsecret-nerc.yaml


joachimweyl:
@knikolla I see the other two ColdFront NERC integration PR's are merged. What are the next steps on this PR to get ColdFront on NERC?

cluster-scope/overlays/nerc-ocp-prod/oauths/cluster_patch.yaml
```diff
@@ -0,0 +1,22 @@
...
+      name: mss-keycloak
+      openID:
+        claims:
+          email:
+            - email
+          name:
+            - name
+          preferredUsername:
+            - preferred_username
+        clientID: ocp-prod
```
naved001, line 17: just wanted to make sure this is what we are calling it instead of nerc-ocp-prod

knikolla: [Correct](https://github.com/nerc-project/mss-keycloak/blob/ca40cfde16e44ecc60a00f369ce9bbaa666448eb/k8s/overlays/nerc-shift-1/clients/ocp-prod.yaml#L37), the client for openstack was just named openstack, and nerc seems reduntant.

naved001 (approved):

cluster-scope/overlays/nerc-ocp-prod/kustomization.yaml
```diff
@@ -13,5 +13,6 @@ resources:
 - apiserver/cluster.yaml
 - secretstores
 
-patches:
-- path: ingresscontrollers/default_patch.yaml
```
knikolla, line 16: By looking at the documentation, it seems that when using `patches` you need to specify both `path` and `target`, which wasn't the case here. So I revised this to just be a strategic merge.

cluster-scope/overlays/nerc-ocp-prod/externalsecrets/oauths-clientsecret-nerc.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: external-secrets.io/v1beta1
+kind: ExternalSecret
+metadata:
+  name: oauths-clientsecret-nerc
+  namespace: openshift-config
+spec:
+  secretStoreRef:
+    name: nerc-cluster-secrets
+    kind: ClusterSecretStore
```
larsks: We're using namespaced secret stores on nerc-ocp-prod (because a cluster-wide secret store could unintentially expose secrets in non-admin namespaces). The simplest change is probably to modify `kind:` here to `SecretStore`, and then we'll need another PR creating a `nerc-cluster-secrets` `SecretStore` in the `openshift-config` namespace.

cluster-scope/overlays/nerc-ocp-prod/kustomization.yaml
```diff
@@ -13,5 +13,6 @@ resources:
...
+patchesStrategicMerge:
+- ingresscontrollers/default_patch.yaml
+- oauths/cluster_patch.yaml
```
larsks: You don't need both `path` and `target` for `patches`. You can just write:

```
patches:
- ingresscontrollers/default_patch.yaml
- oauths/cluster_patch.yaml
```

I prefer this (a single `patches` section) over `patchesStrategicmerge` and `patchesJson6902`.

This is equivalent to the above, but arguably more correct:

```
patches:
- path: ingresscontrollers/default_patch.yaml
- path: oauths/cluster_patch.yaml
```

You only need to provide a `target:` when using JSONpatches (because they don't contain any metadata identifying to which resources they should apply).

larsks (changes requested):

cluster-scope/overlays/nerc-ocp-prod/kustomization.yaml
knikolla: Interesting. Will update. GoLand (IDE) was flagging it as an error and I only did a brief Google search to confirm.

cluster-scope/overlays/nerc-ocp-prod/externalsecrets/oauths-clientsecret-nerc.yaml
knikolla: Makes sense. Are there any changes that we need to do on the Vault side of things to support this?

cluster-scope/overlays/nerc-ocp-prod/kustomization.yaml
joachimweyl: @knikolla is this change ready to be committed?

cluster-scope/overlays/nerc-ocp-prod/externalsecrets/oauths-clientsecret-nerc.yaml
larsks: There are a number of changes we will need to make in the vault configuration (although they are largely the same as we would need to make using a ClusterSecretStore).

joachimweyl:
@knikolla as this is no longer blocked what are the next steps for this?

larsks:
@joachimweyl it's still blocked until we've got the prod cluster minimally configured. There are some pull requests in flight for that right now, and then some work to get external secrets functioning correctly.

joachimweyl:
@larsks I set this as blocked by [Add ocp-prod cluster-scope application to argocd](https://github.com/OCP-on-NERC/nerc-ocp-apps/pull/14) and that is in turn blocked by [Remove ocp-prod clusterversion](https://github.com/OCP-on-NERC/nerc-ocp-config/pull/96) are there any other PR's I should set it blocked by?

cluster-scope/base/config.openshift.io/oauths/cluster/oauth.yaml
```diff
@@ -7,3 +7,5 @@ metadata:
     include.release.openshift.io/single-node-developer: "true"
     release.openshift.io/create-only: "true"
   name: cluster
+spec:
+  identityProviders: []
```
larsks, line 11: I don't think we need this change anymore; this file was updated in ee5f117 to have an empty `spec` which should be suffcient for patching.

cluster-scope/overlays/nerc-ocp-prod/externalsecrets/oauths-clientsecret-nerc.yaml
larsks: This still needs to be updated:

```
spec:
  secretStoreRef:
    kind: SecretStore
    name: nerc-secret-store
```

cluster-scope/overlays/nerc-ocp-prod/kustomization.yaml
larsks: I'd like to see this updated to (a) revert the change to the ingresscontrollers patch and (b) use the same syntax for the `oauths/cluster_patch`.

larsks (changes requested):
I think the dependencies are in place for this, but I think we still need a few minor changes to this PR before it can merge.

joachimweyl:
@larsks since GitHub does not specifically alert you when your CR's are outdated I figured I would let you know that all of your CR's on this PR are now outdated.

larsks (approved):

PR #68, opened 2022-08-25
adds external secrets bundle

chrisstafford:
It looks like nerc-prod was already pulling in the common configs. This PR just adds the config bundle and adds the externalsecrets directory under the ocp-prod overlay.

---
  - https://app.asana.com/0/0/1202467165231357


computate (approved):

naved001 (approved):

cluster-scope/overlays/nerc-ocp-prod/kustomization.yaml
```diff
@@ -5,4 +5,6 @@ commonLabels:
 
 resources:
 - ../common
+- ../../bundles/external-secrets-clustersecretstore
```
larsks: We do not actually want to set up a clustersecretstore on ocp-prod, because this would potentially expose secrets to all cluster users. We're going to want to set up namespaced secret stores where we need them.

larsks (dismissed):

cluster-scope/overlays/nerc-ocp-prod/kustomization.yaml
larsks: ...so if this were to simply pull in `../../bundles/external-secrets` instead it would be fine, and then we would add the namespaced resources in a separate PR.

joachimweyl: @chrisstafford can you take a look at this request?

chrisstafford: I'll take care of this

joachimweyl: @larsks did the implemented fix resolve your CR?

joachimweyl:
@chrisstafford I see 2 approvals. Is this ready for merge?

joachimweyl:
@chrisstafford I see this branch has 2 conflicts, can you please resolve these? Once these conflicts are resolved and @larsks confirms that the CR is resolved is this PR ready to be merged?

larsks:
This PR needs several changes before we can use it, in particular:

- We need serviceaccounts in each namespace that has a namespaced secretstore
- There was a duplicate entry in `kustomization.yaml` for nerc-ocp-prod causing build errors
- The `ClusterSecretStore` was still included in the build

I'm pushing some changes that will correct all of the above.

larsks:
I have re-requested reviews on this since I've made several changes.

knikolla (approved):

joachimweyl:
@naved001 can you please re-review this?

naved001 (approved):

knikolla:
Is this still a draft?

larsks:
@knikolla not anymore :)

PR #74, opened 2022-09-06
Adding namespace to operatorgroups.

computate:


naved001:
My slack comment:
> The way things look right now, I don’t understand why the grafana OperatorGroup is in both namespaces i.e. in openshift-operators-redhat and the grafana namespace.

I would defer to @larsks on this.

computate:
@naved001 @larsks the bad one is the operatorgroup named grafana in the openshift-operators-redhat namespace. I must have created it manually or in an earlier PR before it was refactored into argocd apps.

naved001 (approved):
I think because the grafana argocd app is set in openshift-operators-redhat namespace (https://github.com/OCP-on-NERC/nerc-ocp-apps/blob/main/clusters/nerc-ocp-infra/grafana/application.yaml#L13), so when a namespace was not specified under metadata.namespace for the operatorgroup argocd it ended up creating it in the namespace of the argocd app.

It doesn't hurt to set the expected namespace so I think this PR should be okay.

dystewart (approved):

PR #77, opened 2022-09-07
Add OpenShift Pipelines operator to infra cluster

dystewart:
This PR is to add the OpenShift Pipelines Operator subscription to fix the argoCD sync error we're seeing in argoCD relating to the vault backup job. 

The job uses custom resources such as `task` and `taskRun`, which are added to the cluster through this operator. 

Once merged we should be able to sync the vault resources in argoCD.

---
  - https://app.asana.com/0/0/1202713521265401

### DIFF
--- a/cluster-scope/base/operators.coreos.com/subscriptions/openshift-pipelines-operator/kustomization.yaml
+++ b/cluster-scope/base/operators.coreos.com/subscriptions/openshift-pipelines-operator/kustomization.yaml
@@ -1,0 +1,5 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+namespace: openshift-operators-redhat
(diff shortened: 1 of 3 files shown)


computate (approved):

naved001 (changes requested):
Should we consider disabling the creation of clusterwide RBACs that may be a potential security issue? I know we won't have users on the infra cluster, so it may be okay, but I wanted to run this by folks @jtriley @larsks 

https://docs.openshift.com/container-platform/4.10/cicd/pipelines/installing-pipelines.html#op-disabling-automatic-creation-of-rbac-resources_installing-pipelines

naved001 (approved):

joachimweyl:
@dystewart is this ready for merge?

PR #79, opened 2022-09-08
Adding the patch-operator as an argocd app.

computate:
See the documentation here for the patch operator: https://cloud.redhat.com/blog/enhancing-gitops-with-the-patch-operator

### DIFF
--- a/patch/base/subscriptions/patch-operator/subscription.yaml
+++ b/patch/base/subscriptions/patch-operator/subscription.yaml
@@ -1,0 +1,12 @@
+---
+apiVersion: operators.coreos.com/v1alpha1
+kind: Subscription
+metadata:
+  name: patch-operator
+  namespace: patch-operator
+spec:
+  channel: alpha
+  installPlanApproval: Automatic
+  name: patch-operator
+  source: community-operators
+  sourceNamespace: openshift-marketplace
--- a/patch/base/subscriptions/patch-operator/subscription.yaml
+++ b/patch/base/subscriptions/patch-operator/subscription.yaml
@@ -1,0 +1,13 @@
+---
+apiVersion: operators.coreos.com/v1alpha1
+kind: Subscription
+metadata:
+  name: patch-operator
+  namespace: patch-operator
+spec:
+  channel: alpha
+  installPlanApproval: Automatic
+  name: patch-operator
+  source: community-operators
+  sourceNamespace: openshift-marketplace
+  startingCSV: patch-operator.v0.1.9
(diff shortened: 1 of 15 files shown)


patch/base/subscriptions/patch-operator/subscription.yaml
```diff
@@ -0,0 +1,13 @@
...
+metadata:
+  name: patch-operator
+  namespace: patch-operator
+spec:
+  channel: alpha
+  installPlanApproval: Automatic
+  name: patch-operator
+  source: community-operators
+  sourceNamespace: openshift-marketplace
+  startingCSV: patch-operator.v0.1.9
```
dystewart: Should we omit this startingCSV since we are using automatic installPlan? In Operate First we typically don't include startingCSV with installPlanApproval: Automatic

computate: Good catch @dystewart , I have removed the startingCSV.

larsks (changes requested):
@computate I'd like to request a couple of changes:

1. Please rebase this PR down to a single commit (not that all PRs should be single commits, but two of the commits here have identical message and the third is just a merge).
2. Most operators we're installing right now we're doing via the `cluster-scope/` app. E.g:

    ```
    $ grep -r Subscription cluster-scope/
    cluster-scope/base/operators.coreos.com/subscriptions/acm/subscription.yaml:kind: Subscription
    cluster-scope/base/operators.coreos.com/subscriptions/cert-manager/subscription.yaml:kind: Subscription
    cluster-scope/base/operators.coreos.com/subscriptions/openshift-gitops-operator/subscription.yaml:kind: Subscription
    cluster-scope/base/operators.coreos.com/subscriptions/external-secrets-operator/subscription.yaml:kind: Subscription
    cluster-scope/base/operators.coreos.com/subscriptions/group-sync-operator/subscription.yaml:kind: Subscription
    cluster-scope/base/operators.coreos.com/subscriptions/kubernetes-nmstate-operator/subscriptions.yaml:kind: Subscription
    cluster-scope/base/operators.coreos.com/subscriptions/odf-operator/subscription.yaml:kind: Subscription
    ```

    We should probably treat the Patch operator the same way.

I see that some of the recently added logging/metrics stuff uses this model, so I guess I could be persuaded either way. Generally, we use `cluster-scope` for things that require `cluster-admin` access, and then put namespaced resources in separate app directories.

computate:
I will create a new PR for these changes and close this PR @larsks

PR #87, opened 2022-09-22
Update s3 endpoint ES and vault backup task.yaml to reflect ES change

dystewart:
I believe externalsecrets operator was getting confused because my property and secretKey fields were different in externalsecret.yaml. This PR amends this and updates the referenced key values from the resulting secret in task.yaml.

### DIFF
--- a/vault/overlays/nerc-ocp-infra/backup-job/externalsecret.yaml
+++ b/vault/overlays/nerc-ocp-infra/backup-job/externalsecret.yaml
@@ -13,20 +13,6 @@ spec:
     deletionPolicy: "Retain"
     template:
       engineVersion: v2
-  data:
-  - secretKey: ACCESS_KEY_ID
-    remoteRef:
-      key: nerc/nerc-ocp-infra/vault/backup/s3
-      property: S3AccessKey
-  - secretKey: SECRET_ACCESS_KEY
-    remoteRef:
-      key: nerc/nerc-ocp-infra/vault/backup/s3
-      property: S3SecretKey
-  - secretKey: ENDPOINT
-    remoteRef:
-      key: nerc/nerc-ocp-infra/vault/backup/s3
-      property: S3EndPoint
-  - secretKey: BUCKET
-    remoteRef:
-      key: nerc/nerc-ocp-infra/vault/backup/s3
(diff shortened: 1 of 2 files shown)


vault/overlays/nerc-ocp-infra/backup-job/externalsecret.yaml
```diff
@@ -17,16 +17,16 @@ spec:
   - secretKey: ACCESS_KEY_ID
     remoteRef:
       key: nerc/nerc-ocp-infra/vault/backup/s3
-      property: S3AccessKey
+      property: ACCESS_KEY_ID
```
larsks: Couldn't most of this all be replaced by a single `dataFrom` element?

```
apiVersion: external-secrets.io/v1beta1
kind: ExternalSecret
metadata:
  name: vault-backup-s3-endpoint
spec:
  secretStoreRef:
    name: nerc-cluster-secrets
    kind: ClusterSecretStore
  refreshInterval: "1h"
  target:
    name: vault-backup-s3-endpoint
    creationPolicy: 'Owner'
    deletionPolicy: "Retain"
    template:
      engineVersion: v2
  dataFrom:
    - extract:
        key: nerc/nerc-ocp-infra/vault/backup/s3
```

That gets us:

```
apiVersion: v1
kind: Secret
metadata:
  name: vault-backup-s3-endpoint
type: Opaque
data:
  ACCESS_KEY_ID: ...
  BUCKET: ...
  ENDPOINT: ...
  SECRET_ACCESS_KEY: ...
immutable: false
```

computate: @dystewart It's a great suggetion here. If you do, make sure that the fields in the vault get updated to be the right properties.

computate (approved):

vault/overlays/nerc-ocp-infra/backup-job/externalsecret.yaml
dystewart: @larsks good call I actually forgot that you can pull in all the keys at once like this!

larsks (approved):

naved001 (approved):

PR #88, opened 2022-09-23
Exclude ManagedClusterInfo resources from ArgoCD synchronization

larsks:
When we add a ManagedCluster resource to ACM, ACM generates a `ManagerClusterInfo` resource. This resource copies over any labels from the corresponding ManagedCluster. This means that when we add a ManagedCluster via the config repository, the generated resource inherits the labels added by ArgoCD.

This causes a synchronization conflict, because ArgoCD sees the labels and thinks it should be managing the resource, but no such resource exists in our manifests.

The solution is to add the ManagedClusterInfo resource to the argocd `resource.exclusions` configuration options. This turns out to be tricky, because:

- The `resource.exclusions` is one of many keys in the `argo-cm` configmap.
- The `argocd-cm` configmap also includes the argocd dex configuration
- The dex configuration includes an oidc client secret

So, we can't just generate the config map from the repository. The solution is to use the patch operator and store the dex configuration in the vault.

This requires #68 to merge first (and to operate correctly).


openshift-gitops/overlays/nerc-ocp-infra/patches/dex-config.yaml
```diff
@@ -0,0 +1,22 @@
...
+spec:
+  serviceAccountRef:
+    name: patch-operator
+  patches:
+    dex-config:
+      targetObjectRef:
+        apiVersion: v1
+        kind: ConfigMap
+        name: argocd-cm
+        namespace: lars-sandbox
```
rob-baron: is 'lars-sandbox' where it will actually live nerc-ocp-infra and/or production?

rob-baron (commented):
It looks ok to me, was just curious what the plan is for 'lars-sandbox'.

openshift-gitops/overlays/nerc-ocp-infra/patches/dex-config.yaml
larsks: It is not! That's a typo. Thanks for catching it!

openshift-gitops/overlays/nerc-ocp-infra/patches/dex-config.yaml
```diff
@@ -0,0 +1,22 @@
...
+        kind: ConfigMap
+        name: argocd-cm
+        namespace: openshift-gitops
+      patchTemplate: |
+        data:
+          dex.config: "{{ (index . 1).data.config }}"
+      sourceObjectRefs:
+      - apiVersion: v1
+        kind: Secret
+        name: dex-config
```
naved001, line 21: should this be `dex-client-secret`? that's whats in the ExternalSecret.

larsks: Actually, the externalsecret was wrong. Good catch!

naved001 (approved):

larsks:
NB: This needs to merge before https://github.com/ocp-on-nerc/nerc-ocp-config/issues/68 is useful (because nothing will deploy to nerc-ocp-prod until this merges).

rob-baron (approved):

PR #97, opened 2022-10-04
Disable autosync in generated applications

larsks:
I would like to disable autosync on the nerc-ocp-prod-cluster-scope app, but this isn't possible due to our ArgoCD configuration.

This commit disables autosync in the nerc-cluster-aoa ApplicationSet. This will allow us to control autosync on individual applications via the ArgoCD API/UI.

We can discuss later on whether or not this is something we want to re-enable.


joachimweyl (approved):
Small enough and well described enough that I feel comfortable approving this.

jtriley (approved):

PR #99, opened 2022-10-06
Fix nerc-ocp-prod sync issues

larsks:
This pull request corrects a number of problems that were preventing the nerc-ocp-prod-cluster-scope application from syncing:

- Enable SkipDryRunOnMissingResource for some resources
- Add missing operatorgroup for external secrets
- Fix typo in APIServer certificate configuration
- Base oauth must have a spec


larsks:
Note that I think the solution here for the operator + crd + cr sync issue is a band-aid at best; we're going to need to figure out a more robust way of preventing this problem.

computate (approved):

naved001 (approved):

PR #100, opened 2022-10-07
Increasing the Loki ingestionRate and ingestionRateBurstSize.

computate:
See the Loki Operator configuration: https://loki-operator.dev/docs/api.md/#ingestionlimitspec
And the loki configuration under the hood: https://grafana.com/docs/loki/latest/configuration/#limits_config

### DIFF
--- a/loki/base/lokistacks/lokistack.yaml
+++ b/loki/base/lokistacks/lokistack.yaml
@@ -4,6 +4,11 @@ metadata:
   name: lokistack
   namespace: openshift-operators-redhat
 spec:
+  limits:
+    global:
+      ingestion:
+        ingestionBurstSize: 10
+        ingestionRate: 10
   managementState: Managed
   storage:
     secret:


loki/base/lokistacks/lokistack.yaml
```diff
@@ -4,6 +4,11 @@ metadata:
   name: lokistack
   namespace: openshift-operators-redhat
 spec:
+  limits:
+    global:
+      ingestion:
+        ingestionBurstSize: 10000000
+        ingestionRate: 10000000
```
joachimweyl: if the units are in MB then is this 10TB, do we need that much space?

computate: @joachimweyl Thanks for being more careful than I was checking if it was bytes or mb. I have updated it to: 
```yaml
  limits:
    global:
      ingestion:
        ingestionBurstSize: 10
        ingestionRate: 10
```

joachimweyl (approved):
with new sizing, approved.

larsks (changes requested):
I'd like to see two changes to this pull request:

1. Rebase it down to a single commit.

1. Update the commit message to include an explanation for the change. You could take your message from the slack channel and slightly paraphrase it and I think we would be all set.

computate:
Sure @larsks , I think I'm finally figuring out rebasing my PRs into one commit. The PR is ready for review again, please.

larsks (approved):
Looks good! In the future, please wrap your commit messages at 70 characters. I've update the message in this PR.

PR #103, opened 2022-10-11
Adding custom alerts and Slack integration

computate:
Added a new custom alert named StoragePersistentVolumeFillingUp, for reporting when the PersistentVolumeClaim in a given namespace like openshift-storage is only a small percent free. Slack integration for alerts of a given matcher like "alertname = StoragePersistentVolumeFillingUp" is also setup so that these Alerts show up in the "alerts-nerc-ocp" Slack channel.

See the documentation here: 
- [Configuration | Prometheus](https://prometheus.io/docs/alerting/latest/configuration/)
- [Formatting text for app surfaces | Slack](https://api.slack.com/reference/surfaces/formatting)
- [Create and edit alert rules | Grafana Cloud documentation](https://grafana.com/docs/grafana-cloud/legacy-alerting/grafana-cloud-alerting/create-edit-rules/)
- [ACM Forwarding Alerts](https://access.redhat.com/documentation/en-us/red_hat_advanced_cluster_management_for_kubernetes/2.3/html-single/observability/index#customizing-observability)

### DIFF
--- a/cluster-scope/base/core/configmaps/thanos-ruler-custom-rules/configmap.yaml
+++ b/cluster-scope/base/core/configmaps/thanos-ruler-custom-rules/configmap.yaml
@@ -1,0 +1,75 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: thanos-ruler-custom-rules
+  namespace: open-cluster-management-observability
+data:
+  custom_rules.yaml: |
+    groups:
+      - name: openshift-storage
+        rules:
+
+        - alert: CustomStoragePersistentVolumeFillingUp
+          annotations:
+            summary: PersistentVolume is filling up.
+            description: "The PersistentVolume claimed by {{ $labels.persistentvolumeclaim }} in Namespace {{ $labels.namespace }} is only {{ $value | humanizePercentage }} free."
+          expr: kubelet_volume_stats_available_bytes{persistentvolumeclaim=~"db-noobaa-db-pg-.*|metrics-backing-store-noobaa-pvc-.*|noobaa-default-backing-store-noobaa-pvc-.*"}/kubelet_volume_stats_capacity_bytes < 0.10
+          for: 1m
+          labels:
+            instance: "{{ $labels.instance }}"
+            cluster: "{{ $labels.cluster }}"
+            clusterID: "{{ $labels.clusterID }}"
+            PersistentVolumeClaim: "{{ $labels.persistentvolumeclaim }}"
+            severity: critical
+
+        - alert: CustomStoragePersistentVolumeFillingUpPredicted
+          annotations:
+            summary: PersistentVolume is filling up and is predicted to run out of space in 6h.
+            description: "The PersistentVolume claimed by {{ $labels.persistentvolumeclaim }} in Namespace {{ $labels.namespace }} is only {{ $value | humanizePercentage }} free."
+          expr: (kubelet_volume_stats_available_bytes{persistentvolumeclaim=~"db-noobaa-db-pg-.*|metrics-backing-store-noobaa-pvc-.*|noobaa-default-backing-store-noobaa-pvc-.*"}/kubelet_volume_stats_capacity_bytes) < 0.10 and (predict_linear(kubelet_volume_stats_available_bytes[6h], 4 * 24 * 3600)) < 0
+          for: 1h
+          labels:
+            instance: "{{ $labels.instance }}"
+            cluster: "{{ $labels.cluster }}"
+            clusterID: "{{ $labels.clusterID }}"
+            PersistentVolumeClaim: "{{ $labels.persistentvolumeclaim }}"
+            severity: warning
+
+        - alert: CustomContainerMemoryUsage
+          annotations:
+            summary: Container Memory usage (instance {{ $labels.instance }})
+            description: "Container Memory usage is above 80%\n  VALUE = {{ $value }}\n  LABELS = {{ $labels }}"
+          expr: (sum(container_memory_working_set_bytes{name!=""}) BY (cluster, pod, container) / sum(kube_pod_container_resource_limits{resource="memory"} > 0) BY (cluster, pod, container) * 100) > 80
+          for: 2m
+          labels:
+            instance: "{{ $labels.instance }}"
+            cluster: "{{ $labels.cluster }}"
+            clusterID: "{{ $labels.clusterID }}"
+            severity: warning
+
+        - alert: CustomContainerCpuUsage
+          annotations:
+            summary: Container CPU usage (instance {{ $labels.container_name }})
+            description: "Container CPU usage is above 80%\n  VALUE = {{ $value }}\n  LABELS = {{ $labels }}"
+          expr: sum(avg_over_time(node_namespace_pod_container:container_cpu_usage_seconds_total:sum_rate[1h])) by (cluster, namespace, pod) / avg(kube_pod_container_resource_limits{resource="cpu"}) by (cluster, namespace, pod) * 100 > 80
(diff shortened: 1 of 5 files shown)


computate:
@hpdempsey @joachimweyl @larsks @dystewart @naved001 @OCP-on-NERC/nerc-ops As part of this pull request, I thought I would add a list of alerts provided by operate first for us to decide if they are also relevant to our application and should be added in this PR: 

- `ArgoAppSyncFailed`: For the last minute, ArgoCD applications sync status is not succeeded.The `argocd_app_sync_total` metric is the counter for application sync history. 
- `ArgoCDAppOutOfSync`: Triggered when argocd-application is `OutofSync`
- `PVCFillingUp`: The PersistentVolume claimed by {{ $labels.persistentvolumeclaim}} in Namespace {{ $labels.namespace }} is only {{ $value |humanizePercentage }} free. We are adding a pretty much identical one to this for the openshift-storage namespace. 
- `Probe Success Burn Rate`: Jupyterhub successful probe burn rate  in the last 2 minutes found by Blackbox are failing to meet the 98% SLO. I'm not sure Jupyterhub kind of alerts would apply to our environment. 
- `InstanceDown`: [{{ $labels.instance }}] of job [{{ $labels.job }}] has been down for more than 15 minute. expr: up {namespace="opf-seldon"}== 0. I'm not sure what this one actually does. 
- `ContainerMemoryUsage`: Container Memory usage is above 80%. I think this would cause a lot of noise. 
- `ContainerCpuUsage`: Container CPU usage is above 80%. I think this would cause a lot of noise. 

These are the operate alerts that I found.

computate:
I also updated the alert so it will no longer fire because it will only fire below 10% disk capacity instead of below 99% capacity.

joachimweyl:
@computate is this ready for reviewers to take a look?

computate:
@hpdempsey @joachimweyl @larsks @dystewart @naved001 could you please take a look at the list of Alerts provided by Operate First above in the comments, and also help review the PR and see if we should add any more alerts at this point?

larsks:
For the alerts in your [earlier comment](https://github.com/OCP-on-NERC/nerc-ocp-config/pull/103#issuecomment-1277705033), I agree we don't need the jupyterhub alerts at this time, and for the `InstanceDown` alert it would be nice to do some research to find out what it does and then update the description.

The ContainerMemoryUsage and ContainerCpuUsage alerts might be useful; it would be interesting to set them up and see what sort of alerts we get.

cluster-scope/base/core/configmaps/thanos-ruler-custom-rules/configmap.yaml
```diff
@@ -0,0 +1,34 @@
...
+kind: ConfigMap
+metadata:
+  name: thanos-ruler-custom-rules
+  namespace: open-cluster-management-observability
+data:
+  custom_rules.yaml: |
+    groups:
+      - name: openshift-storage
+        rules:
+        - alert: StoragePersistentVolumeFillingUp
```
larsks: The problem with the `StoragePersistentVolumeFillingUp` alert is that we only care about that *for some volumes*. If you, a user, are running an application on the cluster and you are filling up your PV, we don't need an alert for that -- we only want alerts for things that will impact cluster services or applications that are explicitly maintained by NERC staff.

Can we limit this to alert only on volumes with specific labels (or perhaps only to PVCs in specific namespaces)?

cluster-scope/base/external-secrets.io/externalsecrets/open-cluster-management-observability-alertmanager-config/kustomization.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+commonLabels:
```
larsks: We don't need this commonLabels configuration here; this is applied to all resources in the cluster overlays.

cluster-scope/base/core/configmaps/thanos-ruler-custom-rules/configmap.yaml
computate: Sure @larsks I was able to limit the PVCs with an `expr` like this: 
https://github.com/OCP-on-NERC/nerc-ocp-config/pull/103/files#diff-f62ccb9a4fc420c5eb2adad7f01ffa3346f9c200cc6140dcca0354587f3ab299R16

```
expr: kubelet_volume_stats_available_bytes{persistentvolumeclaim=~"db-noobaa-db-pg-.*|metrics-backing-store-noobaa-pvc-.*|noobaa-default-backing-store-noobaa-pvc-.*"}/kubelet_volume_stats_capacity_bytes < 0.10
```
You can also add a `, namespace="..."` as well.

cluster-scope/base/external-secrets.io/externalsecrets/open-cluster-management-observability-alertmanager-config/kustomization.yaml
computate: Sure @larsks , I have removed it.

cluster-scope/base/external-secrets.io/externalsecrets/open-cluster-management-observability-alertmanager-config/externalsecret.yaml
```diff
@@ -0,0 +1,58 @@
...
+              mrkdwn_in:
+                - text
+                - title
+              text: |{{ .slack_config_text }}
+              title: "{{ .slack_config_title }}"
+              title_link: "{{ .slack_config_title_link }}"
+  data:
+  - secretKey: slack_api_url
+    remoteRef:
+      key: nerc/nerc-ocp-infra/open-cluster-management-observability/slack/alerts-nerc-ocp
```
naved001, line 45: This external secret has a bunch of things that make it environment specific (channel name, the secret). Though we'll probably use the same channel for alerts for both clusters anyway. I would probably put this directly in an overlay but I wouldn't block the PR on that.

naved001 (approved):

cluster-scope/base/core/configmaps/thanos-ruler-custom-rules/configmap.yaml
joachimweyl: @larsks does this resolve the CR?

cluster-scope/base/external-secrets.io/externalsecrets/open-cluster-management-observability-alertmanager-config/kustomization.yaml
joachimweyl: @larsks does this resolve the CR?

cluster-scope/base/external-secrets.io/externalsecrets/open-cluster-management-observability-alertmanager-config/externalsecret.yaml
```diff
@@ -0,0 +1,58 @@
...
+            receiver: default
+            group_by: [cluster, alertname]
+            routes:
+              - receiver: slack-notifications
+                group_by: [alertname, datacenter, app]
+                group_wait: 1d
+                group_interval: 1d
+                repeat_interval: 1d
+                matchers:
+                  - alertname =~ "CustomStoragePersistentVolumeFillingUp|CustomStoragePersistentVolumeFillingUpPredicted|CustomContainerMemoryUsage|CustomContainerCpuUsage"
```
computate, line 29: I just updated the commit to fix the matchers so that it can match all the alerts I setup in the ConfigMap.

larsks (approved):

PR #105, opened 2022-10-12
Create a clusterrole, clusterrolebinding to give `xdmod-reader` servi…

naved001:
…ce account

access to cluster metrics.

### DIFF
--- a/cluster-scope/base/core/namespaces/xdmod/namespace.yaml
+++ b/cluster-scope/base/core/namespaces/xdmod/namespace.yaml
@@ -1,0 +1,5 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: xdmod
+spec: {}
--- a/cluster-scope/base/core/namespaces/xdmod/kustomization.yaml
+++ b/cluster-scope/base/core/namespaces/xdmod/kustomization.yaml
@@ -1,0 +1,4 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+- namespace.yaml
--- a/cluster-scope/base/core/serviceaccounts/xdmod-reader/kustomization.yaml
+++ b/cluster-scope/base/core/serviceaccounts/xdmod-reader/kustomization.yaml
@@ -1,0 +1,5 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+namespace: xdmod
+resources:
+- serviceaccount.yaml
--- a/cluster-scope/base/core/serviceaccounts/xdmod-reader/serviceaccount.yaml
+++ b/cluster-scope/base/core/serviceaccounts/xdmod-reader/serviceaccount.yaml
@@ -1,0 +1,4 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: xdmod-reader
--- a/cluster-scope/base/rbac.authorization.k8s.io/clusterrolebindings/xdmod-reader-xdmod-permissions/clusterrolebinding.yaml
+++ b/cluster-scope/base/rbac.authorization.k8s.io/clusterrolebindings/xdmod-reader-xdmod-permissions/clusterrolebinding.yaml
@@ -1,0 +1,12 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: xdmod-reader-xdmod-permissions
+subjects:
+- kind: ServiceAccount
+  name: xdmod-reader
+  namespace: xdmod
+roleRef:
+  kind: ClusterRole
+  name: xdmod-permissions
+  apiGroup: rbac.authorization.k8s.io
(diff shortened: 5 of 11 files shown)


cluster-scope/overlays/nerc-ocp-prod/clusterrolebindings/xdmod-reader-xdmod-permissions.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: xdmod-reader-xdmod-permissions
+subjects:
+- kind: ServiceAccount
+  name: xdmod-reader
+  namespace: xdmod
```
naved001: @tzumainn I need to know what namespace are we doing things in so we can create one.

cluster-scope/overlays/nerc-ocp-prod/clusterrolebindings/xdmod-reader-xdmod-permissions.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: xdmod-reader-xdmod-permissions
+subjects:
+- kind: ServiceAccount
+  name: xdmod-reader
```
naved001: @larsks should we include the serviceaccount manifest in this repo?

naved001 (commented):
we'll probably need to add @tzumainn to whatever namespace we are doing things in but that'll happen after we have auth working for the prod cluster.

cluster-scope/overlays/nerc-ocp-prod/clusterrolebindings/xdmod-reader-xdmod-permissions.yaml
tzumainn: @naved001 'xdmod' sounds good to me!

computate (approved):

tzumainn (approved):

larsks (changes requested):
In an ideal world, we would move the non-cluster-wide resources (so, the ServiceAccount, the namespaced RoleBinding) into a top-level `xdmod` app...which we're already deploying from https://github.com/CCI-MOC/xdmod-cntr.git.

E.g., the application already deploys the `xdmod` namespace (https://github.com/CCI-MOC/xdmod-cntr/blob/main/k8s/kube-base/ns-xdmod.yaml), so trying to manage it here is going to result in a conflict in argocd.

Alternately, if serviceaccount and clusterrolebindings defined here aren't tied tightly to the `xdmod` app, they should probably be moved into another namespace (and potentially a new top-level app to go along with it).

naved001:
I kind of forgot about  https://github.com/CCI-MOC/xdmod-cntr.git,

> E.g., the application already deploys the xdmod namespace (https://github.com/CCI-MOC/xdmod-cntr/blob/main/k8s/kube-base/ns-xdmod.yaml), so trying to manage it here is going to result in a conflict in argocd.

Looks like that repo is only managing the infra cluster. We will need to do something for the production cluster. I am not sure where we want to manage the namespaced resources for this; a top level `xdmod` app in this repo or with `cci-moc/xdmod-cntr`. 

@larsks dumb question: the namespaced rolebinding in this repo gives the group access to the xdmod namespace, isn't that "privileged" operation and should be deployed by an app with elevated privileges (i.e. cluster-scope)? 

@tzumainn @rob-baron thoughts?

tzumainn:
> I kind of forgot about https://github.com/CCI-MOC/xdmod-cntr.git,
> 
> > E.g., the application already deploys the xdmod namespace (https://github.com/CCI-MOC/xdmod-cntr/blob/main/k8s/kube-base/ns-xdmod.yaml), so trying to manage it here is going to result in a conflict in argocd.
> 
> Looks like that repo is only managing the infra cluster. We will need to do something for the production cluster. I am not sure where we want to manage the namespaced resources for this; a top level `xdmod` app in this repo or with `cci-moc/xdmod-cntr`.
> 
> @larsks dumb question: the namespaced rolebinding in this repo gives the group access to the xdmod namespace, isn't that "privileged" operation and should be deployed by an app with elevated privileges (i.e. cluster-scope)?
> 
> @tzumainn @rob-baron thoughts?

I don't think this account is all that tightly bound with the xdmod application, since all it will do is generate log files. I... could be wrong though; @rob-baron what do you think?

rob-baron:
This service account needs to be located on every openshift cluster that xdmod is expected to be reporting on (likely more than just one).  Furthermore, it should not be within the xdmod namespace as that is reserved for deploying xdmod.

The job that uses the service account tokens need to be able to fetch them at runtime from service like vault(?).

rob-baron:
Mainn,

So how exactly will this work with multiple openshift clusters?

I see you are requesting a service account for this cluster, but what I think you mean is that you need to get a cluster-reader privileges on each openshift cluster - but that isn't what you have here.

cluster-scope/base/core/namespaces/xdmod/namespace.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: xdmod
```
rob-baron, line 4: Change the namespace to xdmod-svcacct (or something similar) so that it doesn't conflict with xdmod

This will also have to be changed elsewhere to be consistent.

rob-baron (changes requested):

cluster-scope/base/core/namespaces/xdmod/namespace.yaml
naved001: I think I will submit another [PR](https://github.com/OCP-on-NERC/nerc-ocp-config/pull/137) and close this one.

PR #106, opened 2022-10-12
Fix logging operator

larsks:
There were some issues with #104 and #70 and a few others that I didn't catch until just now because we're finally deploying this on the prod cluster. This PR makes a number of changes:

- Moves operator subscriptions into the cluster-scope app
- Ensures that the `openshift-operators-redhat` namespace exists
- Reverts #104 because it ends up being unnecessary with this reorganization


computate (approved):

naved001 (approved):

rob-baron (approved):
It should make management of permissions to be easier as open shift-logging should have similar permissions to other operators.

Looks ok to me.

PR #108, opened 2022-10-13
Move subscriptions and operatorgroups from loki into cluster-scope

larsks:
It turns out the `loki/` application is also trying to set up the
operatorgroup for the `openshift-redhat-operators` namespace.

This commit moves both subscriptions and operatorgroups out of loki/ and
into cluster-scope/


computate (approved):

computate (approved):
Looks good, I see we are handling the namespaces for openshift-logging and openshift-operators-redhat, as well as their operatorgroups, and the subscriptions of the Logging and Elasticsearch Operator.

computate (changes requested):
Actually, looks like the build is failing trying to the openshift-operators-redhat operatorgroup, but it's actually found here I think: 

[cluster-scope/base/operators.coreos.com/operatorgroups/redhat-operators/operatorgroup.yaml](https://github.com/OCP-on-NERC/nerc-ocp-config/pull/106/files#diff-02f6abf427b9e4f87d0c67b0718366e23262623e9202119509a8f625ef39517b)

computate (approved):

larsks:
No, I just neglected to include something in my commit :)

cluster-scope/base/operators.coreos.com/operatorgroups/openshift-operators-redhat/operatorgroup.yaml
```diff
@@ -1,6 +1,5 @@
-apiVersion: operators.coreos.com/v1
```
naved001, line 1: don't we need the apiVersion for operatorgroups?

dystewart, line 1: Was removal of apiversion here intentional?

larsks: we do!

naved001 (approved):

cluster-scope/base/operators.coreos.com/operatorgroups/openshift-operators-redhat/operatorgroup.yaml
larsks: It was not!

PR #112, opened 2022-10-17
Patching the secretStoreRef in prod for lokistack-gateway-bearer-token

computate:
The lokistack-gateway-bearer-token ExternalSecret is pointed to a ClusterSecretStore that only exists in the infra cluster. This points the ExternalSecret to the SecretStore available in prod.


dystewart (approved):

larsks (approved):
This looks good! We're also going to need to make a change in the vault configuration to add `openshift-logging` as a valid namespace; I'll get right on that and update you via slack when it's done.

PR #118, opened 2022-10-25
moving all cluster-scoped manifests out of loki application directory

Gregory-Pereira:
Signed-off-by: greg pereira <grpereir@redhat.com>

following discussion in MOC Slack: https://massopencloud.slack.com/archives/C027TDE52TZ/p1666623728336859


dystewart (approved):

naved001 (changes requested):
Thanks @Gregory-Pereira for organizing this! I think we'll need to update the overlay for nerc-ocp-infra to include the 4 things that are now under `cluster-scope/base/` otherwise this change would break the loki deployment that's currently running on nerc-ocp-infra.

joachimweyl:
> we'll need to update the overlay for nerc-ocp-infra to include the 4 things that are now under `cluster-scope/base/` otherwise this change would break the loki deployment that's currently running on nerc-ocp-infra.

@naved001 will this require an additional PR or were you suggesting including it in this PR?

naved001:
@joachimweyl those changes need to be together otherwise it'll break things, so it needs to be part of this PR. We could push a commit to this PR, or close this and submit another PR.

joachimweyl:
> We could push a commit to this PR, or close this and submit another PR.

@Gregory-Pereira can you make the requested updates to this PR?

naved001:
We'll get to it at some point later. For now, I am going to close this PR.

PR #130, opened 2022-10-27
Include group-sync-operator secretstore in nerc-ocp-prod overlay

larsks:


larsks:
Closing in favor of #129

PR #135, opened 2022-10-31
Change mapping method for NERC SSO to Lookup

knikolla:
This ensures that users aren't automatically created if they haven't been provisioned from ColdFront.


joachimweyl (approved):
claim
The default value. Provisions a user with the identity’s preferred user name. Fails if a user with that user name is already mapped to another identity.

lookup
Looks up an existing identity, user identity mapping, and user, but does not automatically provision users or identities. This allows cluster administrators to set up identities and users manually, or using an external process. Using this method requires you to manually provision users. See [Manually Provisioning a User When Using the Lookup Mapping Method](https://docs.openshift.com/container-platform/3.11/install_config/configuring_authentication.html#LookupMappingMethod).

naved001 (approved):

PR #137, opened 2022-11-02
Create a serviceaccount that can read metrics required by xdmod

### LINKED ISSUE #13: Need account to read production metrics
In order for XDMoD to grab the required infomation, I'll need an account that can read those metrics.

naved001:
There's also a clusterrole and a corresponding clusterrolebinding that the serviceaccount is bound to to give it necessary permissions.

To create the service account, a new top level xdmod-reader app is added.

closes https://github.com/OCP-on-NERC/operations/issues/13

The top level app which will create the serviceaccount is here: https://github.com/OCP-on-NERC/nerc-ocp-apps/pull/21

Once the service account is created I will generate a token and place it in vault as suggested by @rob-baron so then it can be used in the infra cluster.

### DIFF
--- a/cluster-scope/base/rbac.authorization.k8s.io/clusterrolebindings/xdmod-reader-permissions/clusterrolebinding.yaml
+++ b/cluster-scope/base/rbac.authorization.k8s.io/clusterrolebindings/xdmod-reader-permissions/clusterrolebinding.yaml
@@ -1,0 +1,12 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: xdmod-reader-permissions
+subjects:
+- kind: ServiceAccount
+  name: xdmod-reader
+  namespace: xdmod-reader
+roleRef:
+  kind: ClusterRole
+  name: xdmod-permissions
+  apiGroup: rbac.authorization.k8s.io
--- a/cluster-scope/base/rbac.authorization.k8s.io/clusterroles/xdmod-permissions/clusterrole.yaml
+++ b/cluster-scope/base/rbac.authorization.k8s.io/clusterroles/xdmod-permissions/clusterrole.yaml
@@ -1,0 +1,12 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: xdmod-permissions
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - namespaces
+  verbs:
+  - get
+  - list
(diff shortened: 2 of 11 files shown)


cluster-scope/base/rbac.authorization.k8s.io/clusterroles/xdmod-permissions/clusterrole.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: xdmod-permissions
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - namespaces
+  verbs:
```
knikolla, line 10: Does xdmod make use of the watch feature as described in https://kubernetes.io/docs/reference/using-api/api-concepts/#efficient-detection-of-changes (which would require adding the watch verb as well)?

cluster-scope/base/rbac.authorization.k8s.io/clusterrolebindings/xdmod-reader-permissions/clusterrolebinding.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: xdmod-reader-permissions
+subjects:
+- kind: ServiceAccount
+  name: xdmod-reader
+  namespace: xdmod
```
knikolla: Should this be `xdmod-reader`?

knikolla (changes requested):

cluster-scope/base/rbac.authorization.k8s.io/clusterroles/xdmod-permissions/clusterrole.yaml
tzumainn: Not that I know of - at least not the part I'm working on that generates the log files. @rob-baron may be able to answer if the main app uses the watch feature.

cluster-scope/base/rbac.authorization.k8s.io/clusterrolebindings/xdmod-reader-permissions/clusterrolebinding.yaml
naved001: yes, it should be. thanks!

naved001: fixed

computate (approved):

knikolla (approved):

cluster-scope/base/rbac.authorization.k8s.io/clusterroles/xdmod-permissions/clusterrole.yaml
rob-baron: xdmod doesn't know anything about kubernetes or openstack.  All knowledge of either platform is contained in the scripts that we (mainn and I are writing).  The watch feature would be an excellent addition to the xdmod-openshift script if it simplifies fetching only the data that has changed.

PR #138, opened 2022-11-03
Fix secret and deployment for acct-mgt

knikolla:
- Since between the initial patches merged, we switched to using a SecretStore rather than a ClusterSecretStore. This patch reworks the external secret to use the namespaced SecretStore and moved the ExternalSecret to the overlay, removing the patch.
- Fix image tag not including `v` in the deployment.


acct-mgt/overlays/nerc-ocp-prod/externalsecret.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: external-secrets.io/v1beta1
+kind: ExternalSecret
+metadata:
+  name: acct-mgt-secrets
+spec:
+  secretStoreRef:
+    name: nerc-secret-store
```
naved001, line 7: you'll need to create this secret store in `acct-mgt` namespace.

Example: https://github.com/OCP-on-NERC/nerc-ocp-config/blob/main/cluster-scope/overlays/nerc-ocp-prod/secretstores/openshift-storage/kustomization.yaml

naved001 (changes requested):

acct-mgt/overlays/nerc-ocp-prod/externalsecret.yaml
knikolla: Ah, thank you, forgot about that.

naved001 (approved):

computate (approved):

PR #139, opened 2022-11-03
Add framework for vault auto-unseal job

dystewart:
This adds the first portion of the cronjob to auto unseal the vault. Once this is merged and confirmed working I'll be adding the actual unsealing functionality.


joachimweyl:
@dystewart do you have the next steps to resolve the Kustomize check failing?

dystewart:
@joachimweyl There's something going on here with the configMap generation I believe. Debugging in a test cluster now

dystewart:
Fixed kustomize build failure and tweaked the seal status script slightly. Most notably the script now writes and updates sealed annotation to the vault statefulset each run

vault/overlays/nerc-ocp-infra/auto-unseal/cronjob.yaml
```diff
@@ -0,0 +1,30 @@
...
+    spec:
+      template:
+        metadata:
+          name: auto-unseal-vault
+        spec:
+          serviceAccountName: auto-unseal-job
+          restartPolicy: Never
+          containers:
+            - name: auto-unseal-vault
+              image: quay.io/dystewar/vault-unseal:latest
```
larsks, line 19: Before merging this change, we need to:

- Move the sources for this image into the ocp-on-nerc repository
- Configure the image to auto-build on quay (or add workflows to build it on github)
- Update this manifest to use the new image location

vault/overlays/nerc-ocp-infra/auto-unseal/cronjob.yaml
```diff
@@ -0,0 +1,30 @@
...
+              imagePullPolicy: IfNotPresent
+              command: ["bash"]
+              args:
+                - /scripts/check-vault-seal.sh
+              volumeMounts:
+                - mountPath: scripts
+                  name: scripts
+          volumes:
```
larsks, line 27: If we're using a custom image already, why are we mounting this script via a configmap rather than just including it in the image?

vault/overlays/nerc-ocp-infra/auto-unseal/scripts/check-vault-seal.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+set -ue
+
+echo "Checking vault seal status..."
+
+oc exec -i nerc-vault-0 -- vault status --format=json > /tmp/sealed.json
```
larsks, line 6: Not really a problem, but by preference I would use `kubectl` instead of `oc` unless you actually need some sort of openshift-specific feature.

larsks (changes requested):
In addition to the inline comments I've left...how does this work? I don't actually see any unsealing anywhere. The script only updates the `sealed` annotation on the statefulset with the current vault status.

vault/overlays/nerc-ocp-infra/auto-unseal/cronjob.yaml
dystewart: @larsks understood, any preference where the sources for the custom image live? If not I will place them in a separate dir under auto-unseal. Also I'll set up a workflow to build the image as well

vault/overlays/nerc-ocp-infra/auto-unseal/cronjob.yaml
dystewart: In the meantime I included the script via cm for clarity, but it certainly makes sense to bake it into the image. Not to mention once the image moves and the sources are added to the repo it'll still be clear anyway. I'll make this change

vault/overlays/nerc-ocp-infra/auto-unseal/scripts/check-vault-seal.sh
dystewart: Noted.. And I agree, will change to kubectl

vault/overlays/nerc-ocp-infra/auto-unseal/cronjob.yaml
naved001: what's in the custom image?

Is the correct dckerfile to look at?: https://github.com/dystewart/k8s-vault-auto-unseal/blob/main/docker/Dockerfile

larsks: Hosting the sources for the custom image in this repository seems ideal.

dystewart:
@larsks great feedback here ty. 

> In addition to the inline comments I've left...how does this work? I don't actually see any unsealing anywhere. The script only updates the `sealed` annotation on the statefulset with the current vault status.

@larsks My initial plan was to use this annotation that's being added to the statefulset to trigger a separate unseal script... But coming back to this I think it makes far more sense to add logic to this job to unseal if necessary. 

There is still value to adding the sealed annotation to the statefulset imo, and I placed it there to potentially be monitored/log when we are sealed, I'll have to work with @computate for this though

I'll make the changes requested above and rework the job to unseal here as well

vault/overlays/nerc-ocp-infra/auto-unseal/cronjob.yaml
joachimweyl: @dystewart what is your plan for resolving this CR?

joachimweyl:
@dystewart checking-in on the status of this PR.

joachimweyl:
@dystewart can we get an update on this PR?

dystewart:
@joachimweyl haven't had a chance to pick this one up in a while, and when last working on it the job was still in a rather rudimentary state. I still need to add the logic for actually unsealing the vault when it's found sealed.
And probably will want some type of logging to keep track of that

joachimweyl:
@dystewart is this still low on priorities?

PR #145, opened 2022-11-11
Add external-apps-ingress-certificate.yaml ES to prod

dystewart:
The external ingressController in prod is looking for the cert which was not created in the cluster, since it was not added to the kustomization file.
This should allow the ingress to create successfully. 

In response to: [https://github.com/OCP-on-NERC/operations/issues/29](https://github.com/OCP-on-NERC/operations/issues/29)

### DIFF
--- a/cluster-scope/overlays/nerc-ocp-prod/externalsecrets/kustomization.yaml
+++ b/cluster-scope/overlays/nerc-ocp-prod/externalsecrets/kustomization.yaml
@@ -3,6 +3,7 @@ kind: Kustomization
 resources:
 - default-ingress-certificate.yaml
 - default-api-certificate.yaml
+- external-apps-ingress-certificate.yaml
 - oauths-clientsecret-nerc.yaml
 - github-client-secret.yaml
 - github-group-sync-token.yaml


computate (approved):

cluster-scope/overlays/nerc-ocp-prod/externalsecrets/kustomization.yaml
```diff
@@ -3,6 +3,7 @@ kind: Kustomization
 resources:
 - default-ingress-certificate.yaml
 - default-api-certificate.yaml
+- external-apps-ingress-certificate.yaml
```
larsks, line 6: That file references:

```
  dataFrom:
    - extract:
        key: nerc/nerc-ocp-prod/openshift-ingress/external-apps-ingress-certificate
```

But no such key exists in the vault:

```
$ vault kv  list nerc/nerc-ocp-prod/openshift-ingress/
Keys
----
default-ingress-certificate
route53-creds
```

I'm checking to see if we have an existing certificate.

With access via the public network now working, we should be able to configure a letsencrypt-based secretstore for acquiring public certificates.

larsks (changes requested):

cluster-scope/overlays/nerc-ocp-prod/externalsecrets/kustomization.yaml
dystewart: @larsks Ah okay, that would make this very unhelpful in that case. All of this must have been set up as more of a layout than a working implementation. Let me know what we need here and if that cert exists

PR #150, opened 2022-11-15
Add openshift-storage and openshift-gitops namespaces to prod

dystewart:
This should make several of the errors in prod argocd happy


larsks:
Right, but I don't think adding the `openshift-gitops` namespace is necessarily the right thing to do -- I think the fact that things are targeting that namespace represents a configuration issue that we should resolve.

The openshift-storage business should be taken care of when we re-enable ODF.

larsks (changes requested):
Right, but I don't think adding the `openshift-gitops` namespace is necessarily the right thing to do -- I think the fact that things are targeting that namespace represents a configuration issue that we should resolve.

The openshift-storage business should be taken care of when we re-enable ODF.

larsks (approved):

naved001 (approved):

PR #151, opened 2022-11-15
Move all ODF-related resources into feature/odf

larsks:
This allows us to enable/disable ODF with a single comment.

With this commit, ODF is still disabled.


cluster-scope/overlays/nerc-ocp-prod/kustomization.yaml
```diff
@@ -7,8 +7,8 @@ resources:
 - ../common
 - ../../bundles/acct-mgt
 - ../../bundles/metallb
-# - ../../bundles/odf-external
 - ../../bundles/xdmod-reader
+# feature/odf
```
dystewart, line 11: Remove comment

cluster-scope/overlays/nerc-ocp-prod/feature/odf/kustomization.yaml
```diff
@@ -0,0 +1,11 @@
...
+kind: Kustomization
+commonLabels:
+  nerc.mghpcc.org/feature: odf
+
+components:
+  - ../../../../components/nerc-secret-store
+
+resources:
+  - ../../../../bundles/odf-external
+  - externalsecrets/rook-ceph-external-cluster-details.yaml
```
naved001, line 11: we should add the storageclass patch to this (it is commented out right now, so wouldn't matter).

cluster-scope/overlays/nerc-ocp-prod/kustomization.yaml
larsks: That comment was intentional :)

cluster-scope/overlays/nerc-ocp-prod/feature/odf/kustomization.yaml
larsks: Good point, I've updated the pull request.

dystewart (approved):

naved001 (approved):

PR #155, opened 2022-11-16
Correct issues with metallb configuration

larsks:
This PR fixes aspects of the metallb configuration that were using syntax only available in 4.11, and it also disabled address auto-assignment.

### DIFF
--- a/cluster-scope/overlays/nerc-ocp-prod/metallb/kustomization.yaml
+++ b/cluster-scope/overlays/nerc-ocp-prod/metallb/kustomization.yaml
@@ -2,6 +2,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 namespace: metallb-system
 resources:
-- ipaddresspools/public.yaml
-- l2advertisements/default.yaml
+- addresspools/public.yaml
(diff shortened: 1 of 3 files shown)


dystewart (approved):

cluster-scope/overlays/nerc-ocp-prod/metallb/kustomization.yaml
```diff
@@ -2,6 +2,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 namespace: metallb-system
 resources:
-- ipaddresspools/public.yaml
-- l2advertisements/default.yaml
```
naved001, line 6: do you want to remove this file too?

larsks: I was thinking of keeping them for when we upgrade to 4.11 at some point, but I can go ahead and remove them.

larsks: Deleted.

naved001 (approved):

PR #157, opened 2022-11-18
Configure ingressController to use routeSelector and configure external ingress service

dystewart:


computate (approved):

cluster-scope/overlays/nerc-ocp-prod/services/router-external-apps-ingress-controller.yaml
```diff
@@ -0,0 +1,33 @@
...
+  annotations:
+    metallb.universe.tf/address-pool: public
+    traffic-policy.network.alpha.openshift.io/local-with-fallback: ""
+  labels:
+    app: router
+    ingresscontroller.operator.openshift.io/owning-ingresscontroller: external-apps-ingress-controller
+    router: router-external-apps-ingress-controller
+  name: router-external-apps-ingress-controller
+  namespace: openshift-ingress
+spec:
```
larsks, line 13: We need to add the `domain` option to this configuration; see [the ingress docs](https://docs.openshift.com/container-platform/4.10/networking/configuring_ingress_cluster_traffic/configuring-ingress-cluster-traffic-ingress-controller.html#nw-ingress-sharding-route-labels_configuring-ingress-cluster-traffic-ingress-controller).

The domain should be `apps.shift.nerc.mghpcc.org`.

larsks (dismissed):
I've noted one additional change we need to make here.

Since you're going to be making changes *anyway*, I would suggest making the commit messages slightly more descriptive:

- For the "Change ingressController to use routeSelector", consider adding to the commit body our reasons for that change.

- For "Configure address for external ingress service", it's probably worth noting which fields in particular we're adding with this change (and why), since otherwise it's not obvious.

In either or both cases, consider adding a link to the relevant documentation if you think it's appropriate (see e.g. 480edd0058553419819257c6b01f8394a9bd2054 for an example of how I typically do that).

cluster-scope/overlays/nerc-ocp-prod/ingresscontrollers/external-apps-ingress-controller.yaml
```diff
@@ -15,6 +15,6 @@ spec:
     nodeSelector:
       matchLabels:
         nerc.mghpcc.org/external-ingress: 'true'
-  namespaceSelector:
+  routeSelector:
```
larsks, line 18: We need to add the `domain` option **here** (ignore that earlier comment if you saw it; I've remove it). Relevant docs are https://docs.openshift.com/container-platform/4.10/networking/configuring_ingress_cluster_traffic/configuring-ingress-cluster-traffic-ingress-controller.html#nw-ingress-sharding-route-labels_configuring-ingress-cluster-traffic-ingress-controller.

We should set domain to `apps.shift.nerc.mghpcc.org`.

larsks:
I've removed my requested change (apparently I am unable to read), but I would still like to see the commit messages updated. Sorry for the noise!

cluster-scope/overlays/nerc-ocp-prod/ingresscontrollers/external-apps-ingress-controller.yaml
```diff
@@ -15,6 +15,6 @@ spec:
...
     matchLabels:
       type: external
```
larsks: Actually, here is one more change I think we should make. I think that `type` is probably too generic. Let's (a) pick a more descriptive name, and (b) let's namespace it like we have for other locally relevant labels.

How about:

```
routerSelector:
  matchLabels:
    nerc.mghpcc.org/external-ingress: "true"
```

This is the same label we use to identify nodes that host the external ingress address (and placement for the external ingress service).

dystewart: @larsks good points, updated the commit messages and changed the label

larsks:
@dystewart this looks fine, but there are apparently still missing pieces. See https://github.com/OCP-on-NERC/operations/issues/41 for some details. I'd like to hold off merging this until we get things working (because we'll probably need to update the same set of files).

naved001 (approved):
looks good. I don't know a lot about patch operator, but is there a reason why we aren't using that?

larsks:
> looks good. I don't know a lot about patch operator, but is there a reason why we aren't using that?

I'm a little leery of it: it appears to be written by a single person, and I disagree with the security model (see e.g. https://github.com/redhat-cop/patch-operator/issues/47).

PR #159, opened 2022-11-22
Add secretstore for openshift-ingress-operator namespace

larsks:
This enables us to fetch the external apps certificate from the vault.


larsks:
Wait, this is the wrong solution. We should be getting cert-manager set up for this.

PR #160, opened 2022-11-22
fix/external ingress cert

larsks:
- Remove unused externalsecret
- Configure certificate issuer for external ingress operator


larsks:
Marking this as draft for a moment -- some of these resources need to be in the `openshift-ingress` namespace instead of `openshift-ingress-operator`.

larsks:
Now things are in the right places.

dystewart (approved):

naved001 (approved):

PR #167, opened 2022-12-01
Allow nerc-ops to control application sync

### LINKED ISSUE #43: Allow nerc-ops users to modify argocd applications
When logged in via GitHub, we have read-only permissions by default and must use impersonation to make changes to the cluster. This works fine from the CLI or within the main openshift console UI, but it is problematic when interacting with argocd: we would like nerc-ops members to be able to manage the "auto sync" setting for applications without requiring elevated privileges.

Investigating and apply the necessary RBAC to allow nerc-ops members to manage argocd application sync policy.

larsks:
Update the argocd rbac policy so that members of the nerc-ops group can
control application sync settings via the web ui (or the argocd cli).

Closes ocp-on-nerc/operations#43

### DIFF
--- a/openshift-gitops/overlays/nerc-ocp-infra/patches/argocds/openshift-gitops.yaml
+++ b/openshift-gitops/overlays/nerc-ocp-infra/patches/argocds/openshift-gitops.yaml
@@ -18,3 +18,12 @@ spec:
       - "ManagedClusterInfo"
       clusters:
       - "*"
+  rbac:
+    policy: |
+      p, role:operator, applications, sync, */*, allow
+      p, role:operator, applications, update, */*, allow
+
+      g, system:cluster-admins, role:admin
+      g, cluster-admins, role:admin
+      g, nerc-ops, role:operator
+    scopes: '[groups]'


openshift-gitops/overlays/nerc-ocp-infra/patches/argocds/openshift-gitops.yaml
```diff
@@ -18,3 +18,12 @@ spec:
       - "ManagedClusterInfo"
       clusters:
       - "*"
+  rbac:
```
naved001, line 21: The linked document says to create a `ConfigMap` called `argocd-rbac-cm` and that would include the policy. How did you figure out we can include `rbac` in this resource of kind `ArgoCD`? I can't seem to find this in the documents.

openshift-gitops/overlays/nerc-ocp-infra/patches/argocds/openshift-gitops.yaml
```diff
@@ -18,3 +18,12 @@ spec:
...
+    policy: |
+      p, role:operator, applications, sync, */*, allow
+      p, role:operator, applications, update, */*, allow
+
+      g, system:cluster-admins, role:admin
+      g, cluster-admins, role:admin
+      g, nerc-ops, role:operator
```
naved001, line 28: the linked argocd document examples have this as `your-github-org:your-team` but I guess since we are syncing the groups we don't specify the github org? Just trying to understand. 

Also what does the `g` mean? The argocd document used this in examples but didn't explain what it is.

larsks: > the linked argocd document examples have this as your-github-org:your-team but I guess since we are syncing the groups we don't specify the github org?

Right, we're simply referencing openshift groups in this policy.

> Also what does the g mean? The argocd document used this in examples but didn't explain what it is.

I noticed that lack of documentation as well!  I've decided the `p` means `policy` and `g` means `group`. In fact, with a quick search just now I found [this example](https://github.com/argoproj/argo-cd/blob/master/docs/operator-manual/argocd-rbac-cm.yaml) which seems to confirm that interpretation:

```
# Policy rules are in the form:
#   p, subject, resource, action, object, effect
# Role definitions and bindings are in the form:
#   g, subject, inherited-subject
```

openshift-gitops/overlays/nerc-ocp-infra/patches/argocds/openshift-gitops.yaml
larsks: You need to look at the docs for the [ArgoCD operator](https://argocd-operator.readthedocs.io/en/latest/reference/argocd/) or for the [OpenShift GitOps operatpr](https://docs.openshift.com/container-platform/4.10/cicd/gitops/argo-cd-custom-resource-properties.html).

naved001 (approved):

computate (approved):

PR #171, opened 2022-12-07
Configure ArgoCD to use the "annotation" resource tracking method

larsks:
Set the argocd `resourceTrackingMethod` to `annotation`. This tells ArgoCD to track resources using a generated tracking id, rather than relying on the `app.kubernetes.io/instance` label. From the docs [1]:

> The advantages of using the tracking id annotation is that there are no
> clashes any more with other Kubernetes tools and Argo CD is never
> confused about the owner of a resource.

[1]: https://argo-cd.readthedocs.io/en/stable/user-guide/resource_tracking/


larsks:
I'm going to retire this one for now. I'd like to test it somewhere first.

PR #172, opened 2022-12-09
Label all bundle generated resources

larsks:
This allows us to identify the resources generated by a bundle, and can
also be used if you want to install a single bundle while still applying
any customizations from a cluster overlay:

    cd cluster-scope/overlays/nerc-ocp-prod
    kustomize build |
      yq 'select(.metadata.labels."nerc.mghpcc.org/bundle" == "metallb")' |
      kubectl apply -f-


dystewart (approved):
Looks good, I like this addition

naved001 (approved):

PR #178, opened 2023-01-06
Manage node labels using patch operator

### LINKED ISSUE #50: Manage node labels via our configuration repository
Various aspects of our OpenShift environment depend on labels being applied to Node resources.

We are currently [manually applying these labels](https://github.com/OCP-on-NERC/operations/issues/28#issuecomment-1317738000). We would like to manage node labels through our gitops workflow, both so that we have the change history and so that when we rebuild a cluster the appropriate labels are re-applied to nodes without requiring manual intervention.

naved001:
closes https://github.com/OCP-on-NERC/operations/issues/50


cluster-scope/overlays/nerc-ocp-infra/redhatcop.redhat.io/odf-node-patcher.yaml
```diff
@@ -0,0 +1,60 @@
...
+  namespace: openshift-storage
+spec:
+  serviceAccountRef:
+    name: patcher
+  patches:
+    odf-node-label-ctl-0:
+      targetObjectRef:
+        apiVersion: v1
+        kind: Node
+        name: ctl-0
```
larsks: I think that if we drop the `name` field from the `targetObjectRef`, the patch would apply to all node resources. There's a table in the [patch enforcement](https://github.com/redhat-cop/patch-operator#runtime-patch-enforcement) section of the README that suggests this.

larsks (changes requested):

cluster-scope/overlays/nerc-ocp-infra/redhatcop.redhat.io/odf-node-patcher.yaml
naved001: makes sense, I have made that change for the nerc-ocp-infra nodes.

computate (approved):

cluster-scope/overlays/nerc-ocp-prod/feature/odf/redhatcop.redhat.io/odf-node-patcher.yaml
```diff
@@ -0,0 +1,200 @@
...
+apiVersion: redhatcop.redhat.io/v1alpha1
+kind: Patch
+metadata:
+  name: odf-node-label-wrk-10
+  namespace: openshift-storage
+spec:
+  serviceAccountRef:
+    name: patcher
+  patches:
+    odf-node-label-wrk-10:
```
larsks: Why are we only applying the `openshift-storage` label to a subset of the nodes on ocp-prod?

naved001: The patch was for all worker nodes actually, after our discussion I have updated this to apply to all worker nodes by matching the Kubernetes worker label.

cluster-scope/overlays/nerc-ocp-prod/feature/odf/redhatcop.redhat.io/odf-node-patcher.yaml
```diff
@@ -1,198 +1,20 @@
...
-  patches:
-    odf-node-label-wrk-9:
-      targetObjectRef:
-        apiVersion: v1
-        kind: Node
-        name: wrk-9
+        labelSelector:
+          matchExpressions:
+            - key: node-role.kubernetes.io/worker
+              operator: Exists
```
naved001: I couldn't find an example in their documentation on how to do this, so this is based off of the kubernetes crd documentation and the server dry run said this was valid.

larsks: We can run `k explain patch.spec.patches.targetObjectRef.labelSelector` to get the API docs for that field, which tell us that it's a standard label selector -- which means that it supports both `matchLabels` and `matchExpressions`:

```
$ kubectl explain patch.spec.patches.targetObjectRef.labelSelector
KIND:     Patch
VERSION:  redhatcop.redhat.io/v1alpha1

RESOURCE: labelSelector <Object>

DESCRIPTION:
     LabelSelector selects objects by label

FIELDS:
   matchExpressions     <[]Object>
     matchExpressions is a list of label selector requirements. The requirements
     are ANDed.

   matchLabels  <map[string]string>
     matchLabels is a map of {key,value} pairs. A single {key,value} in the
     matchLabels map is equivalent to an element of matchExpressions, whose key
     field is "key", the operator is "In", and the values array contains only
     "value". The requirements are ANDed.
```

So  your syntax looks correct, but we could also write:

```
labelSelector:
  matchLabels:
    node-role.kubernetes.io/worker: ""
```

naved001: Thank you for explaining this. I was trying to read through the output of kubectl describe crd patch, but this is much more readable. I have updated it to use `matchLabels` now.

larsks (approved):

PR #183, opened 2023-01-11
Fix namespace for the patch of openshift-ingress

naved001:
~it matters because the patch will use the serviceaccount in that namespace which has the permissions to patch `Node` resources.~
The odf serviceaccount patcher should also have worked, but it's better this  stay in the correct namespace.


naved001:
P.S the changes in the overlay for nerc-ocp-prod didn't apply. I suspect something to do with argocd because of the ongoing cluster rebuilds.

computate (approved):

dystewart (approved):

PR #190, opened 2023-01-26
Including missing observability custom metrics and updated loki ca bundle

computate:
Two fixes: 
- The observability-metrics-custom-allowlist was missing after the rebuild
- The custom CA Bundle that needs to be configured in a configmap needed updating after a cluster rebuild.


jtriley (approved):

naved001 (approved):

larsks:
@computate Instead of using this custom CA bundle, can we just use the CA certificates provided automatically by our service account? These are mounted in `/run/secrets/kubernetes.io/serviceaccount/`. If we do this, we wouldn't need to reconfigure anything when (a) reinstalling the cluster or (b) deploying on a new cluster.

computate:
Looks like you're right @larsks , not sure why I didn't see this before, but it looks like there is a configmap `openshift-service-ca.crt` in every namespace that I can point to in the LokiStack instead.

computate:
I added a followup PR to point to the default openshift-service-ca.crt: https://github.com/OCP-on-NERC/nerc-ocp-config/pull/191

PR #191, opened 2023-01-26
Updating the lokistack-ca-bundle-custom configmap to the default openshift-service-ca.crt

computate:


jtriley (approved):

larsks (approved):
This looks fine. I'm going to approve it, but if you're feeling motivated I would appreciate a slightly more robust commit message (what is this setting used for? Where does openshift-service-ca.crt come from? Yes, **I** know and **you** know the answers to those questions, but I like the commit history to be accessible to someone who may be less familiar with things.)

PR #196, opened 2023-02-22
Upgrading Loki Operator, removing custom grafana, adding logging-view-plugin

computate:
In order to apply retention to our Loki logs, we need to upgrade to the new Loki Operator 5.6. Loki 5.6 deprecates querying logs directly from the querier and applies different security to provide the logs in OpenShift instead. We are removing grafana and adding the logging-view-plugin to the OpenShift console instead. We are also moving Loki to the recommended openshift-logging namespace instead.


computate:
Currently there is a feature request to add better control over groups, roles, and logs besides cluster-admin: 
https://issues.redhat.com/browse/RFE-3610

dystewart (approved):

computate:
The nice advantage of these updates is that the whole Logging configuration is now Red Hat Supported, instead of using a custom grafana instance. 
- Here are the docs related to [OpenShift Cluster Logging with Loki](https://docs.openshift.com/container-platform/4.12/logging/cluster-logging-loki.html)
- Here are the docs for [Forwarding audit logs to the log store](https://docs.openshift.com/container-platform/4.12/logging/config/cluster-logging-log-store.html#cluster-logging-elasticsearch-audit_cluster-logging-store)

naved001:
> we need to upgrade to the new Loki Operator 5.6

> We are removing grafana and adding the logging-view-plugin to the OpenShift console instead

> We are also moving Loki to the recommended openshift-logging namespace instead.

Is it possible to split this PR into different commits? It would make it easier to review and understand what's going on. Thanks!

computate:
@naved001 @larsks @joachimweyl Thanks for the feedback. I have split this PR up into separate commits, please review again.

naved001 (approved):
@computate Thank you!

PR #198, opened 2023-02-24
Setting up test cluster overlay 

dystewart:
In this PR I'm setting up external ceph storage and adding external-secrets operator


dystewart:
Added odf-operator bundle to nerc overlay

PR #200, opened 2023-03-03
Expose console and API externally

### LINKED ISSUE #66: prod cluster API server needs to be externally accessible
Currently users would need to be on VPN in order to interact with the API server. For production we'll need to be able to access the API server from outside VPN using the `api.shift.nerc.mghpcc.org` address.

### LINKED ISSUE #67: prod cluster OpenShift console needs to be externally accessible
Users need to be able to access the OpenShift console externally. If we apply the external route label to the existing openshift-console route that would give us:

https://console-openshift-console.apps.shift.nerc.mghpcc.org

We might want to consider adding a custom route to the console to shorten this to just `https://shift.nerc.mghpcc.org`

larsks:
This commit includes several changes to make the OpenShift console and API
available externally (without VPN access). Broadly, there are

- Modify the cluster-ingress configuration to use externally visible
  hostnames for the console, downloads, and oauth services
- Add explicit external routes (i.e, routes with a label that matches
  our external ingress service) for these services
- Add a proxy to the OpenShift API, exposed with an external route

With these changes in place, it is possible to access the console from
outside the VPN. Some URLs (e.g., when using the "copy login command" menu
option) will still use the internal, vpn-only hostnames.

Closes ocp-on-nerc/operations#67
Closes ocp-on-nerc/operations#66

---
  - https://app.asana.com/0/0/1204165599358399

### DIFF
--- a/cluster-scope/overlays/nerc-ocp-prod/feature/external-console-access/api-proxy/deployment.yaml
+++ b/cluster-scope/overlays/nerc-ocp-prod/feature/external-console-access/api-proxy/deployment.yaml
@@ -1,0 +1,29 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: api-proxy
+spec:
+  replicas: 3
+  template:
+    spec:
+      serviceAccountName: haproxy
+      containers:
+      - name: haproxy
+        image: docker.io/haproxy:lts
+        ports:
+        - containerPort: 8080
+          name: http
+        volumeMounts:
+          - name: haproxy-config
+            mountPath: /usr/local/etc/haproxy
+          - name: haproxy-state
+            mountPath: /var/lib/haproxy
+        securityContext:
+          allowPrivilegeEscalation: false
+          runAsUser: 0
+      volumes:
+        - name: haproxy-config
+          configMap:
+            name: haproxy-config
+        - name: haproxy-state
+          emptyDir: {}
--- a/cluster-scope/overlays/nerc-ocp-prod/feature/external-console-access/api-proxy/role.yaml
+++ b/cluster-scope/overlays/nerc-ocp-prod/feature/external-console-access/api-proxy/role.yaml
@@ -1,0 +1,13 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: haproxy-run-as-root
+rules:
+- apiGroups:
+  - security.openshift.io
+  resourceNames:
+  - anyuid
+  resources:
+  - securitycontextconstraints
+  verbs:
+  - use
--- a/cluster-scope/overlays/nerc-ocp-prod/feature/external-console-access/routes/api.yaml
+++ b/cluster-scope/overlays/nerc-ocp-prod/feature/external-console-access/routes/api.yaml
@@ -1,0 +1,17 @@
+apiVersion: route.openshift.io/v1
+kind: Route
+metadata:
+  name: api-external
+  namespace: openshift-kube-apiserver
+  labels:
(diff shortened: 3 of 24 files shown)


cluster-scope/overlays/nerc-ocp-prod/feature/external-console-access/api-proxy/deployment.yaml
```diff
@@ -0,0 +1,29 @@
...
+        - containerPort: 8080
+          name: http
+        volumeMounts:
+          - name: haproxy-config
+            mountPath: /usr/local/etc/haproxy
+          - name: haproxy-state
+            mountPath: /var/lib/haproxy
+        securityContext:
+          allowPrivilegeEscalation: false
+          runAsUser: 0
```
knikolla, line 23: Why does this need to run as root?

cluster-scope/overlays/nerc-ocp-prod/feature/external-console-access/api-proxy/role.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
```
naved001, line 2: Didn't know what this role was doing, and after some looking around I landed here: 
https://docs.openshift.com/container-platform/4.10/authentication/managing-security-context-constraints.html#role-based-access-to-ssc_configuring-internal-oauth

Question: Can non-admins create a role like this in their namespace and run their containers as root?

cluster-scope/overlays/nerc-ocp-prod/feature/external-console-access/routes/api.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: route.openshift.io/v1
+kind: Route
+metadata:
+  name: api-external
```
naved001, line 4: I am confused, there's a route (cluster-scope/overlays/nerc-ocp-prod/feature/external-console-access/api-proxy/route.yaml) with the same name but to the service `api-proxy`. This one is to `apiserver` Service. Both have the same host though `api.apps.shift.nerc.mghpcc.org`.

Additionally, this is not included in the kustomization file in this directory.

cluster-scope/overlays/nerc-ocp-prod/feature/external-console-access/api-proxy/deployment.yaml
larsks: When I wrote this, I assumed that haproxy's default behavior of start as root and then suid-ing to an unprivileged user was intrinsic to haproxy, but it turns out that removing the `user` and `group` directories is sufficient.

cluster-scope/overlays/nerc-ocp-prod/feature/external-console-access/api-proxy/role.yaml
larsks: @naved001 if someone tries to create a rolebinding that grants permissions they don't already have, it will fail with:

```
Error from server (Forbidden): error when creating ".": roles.rbac.authorization.k8s.io "allow-anyuid" is forbidden: user "testuser" (groups=["system:authenticated:oauth" "system:authenticated"]) is attempting to grant RBAC permissions not currently held:
{APIGroups:["security.openshift.io"], Resources:["securitycontextconstraints"], ResourceNames:["anyuid"], Verbs:["use"]}
Error from server (NotFound): error when creating ".": roles.rbac.authorization.k8s.io "allow-anyuid" not found
```

joachimweyl:
This was an example of one way to do this, we found a better way

PR #207, opened 2023-03-13
Feature/external lb

larsks:
This pull request contains the changes necessary to move from the sharded ingress model to using an external load balancer to expose the cluster externally.

~~Marked as a draft because I am still working on it.~~

There are several commits in this PR; I've tried to structure them logically so that you can review one commit at a time.


jtriley (commented):
Should we also remove this and update the kustomization.yaml given that we're not handling public ingress on the cluster anymore?

- cluster-scope/overlays/nerc-ocp-prod/nodenetworkconfigurationpolicies/vlan-2180-external.yaml

larsks:
> Should we also remove this and update the kustomization.yaml given that we're not handling public ingress on the cluster anymore?

I had some thought in my head that we may still want to make use of metallb at some point in the future, but it will probably simplify things to remove this for now and put it back if we need it.

cluster-scope/overlays/nerc-ocp-prod/certificates/default-ingress-certificate.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: cert-manager.io/v1
+kind: Certificate
+metadata:
+  name: default-ingress-certificate
+  namespace: openshift-ingress
+spec:
+  issuerRef:
+    name: letsencrypt-staging-dns01
```
larsks, line 8: Note that I have intentionally left this configured with the staging server for now; once cert generation has been verified to work correctly we'll swap these to the production issuers.

cluster-scope/overlays/nerc-ocp-prod/certificates/default-api-certificate.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: cert-manager.io/v1
+kind: Certificate
+metadata:
+  name: default-api-certificate
+  namespace: openshift-config
+spec:
+  issuerRef:
+    name: letsencrypt-staging-dns01
```
larsks, line 8: (See comment below re: staging vs production issuers)

cluster-scope/overlays/nerc-ocp-prod/kustomization.yaml
```diff
@@ -6,56 +6,47 @@ commonLabels:
...
-    name: ".*"
-  path: patches/skipdryrunonmissingresource.yaml
-
-- target:
-    group: "redhatcop.redhat.io"
-    name: ".*"
-  path: patches/skipdryrunonmissingresource.yaml
+- path: patches/skipdryrunonmissingresource.yaml
+  target:
+    group: metallb.io
```
jtriley, line 35: Likely harmless but if we're removing metallb do we need this patch?

computate (approved):

cluster-scope/overlays/nerc-ocp-prod/kustomization.yaml
larsks: It is harmless, it is not necessary...but I guess I'm inclined to leave it here in case we re-introduce metallb at some point.

jtriley (approved):

dystewart (approved):

naved001 (approved):

PR #208, opened 2023-03-14
Patch nerc-secret-store in test cluster overlay

dystewart:
Patches the nerc-secret-store in the test cluster to use test cluster k8s auth method in vault

### DIFF
--- a/cluster-scope/overlays/nerc-ocp-test/feature/odf/kustomization.yaml
+++ b/cluster-scope/overlays/nerc-ocp-test/feature/odf/kustomization.yaml
@@ -15,3 +15,4 @@ resources:
 
 patches:
   - path: storageclasses/ocs-external-storagecluster-ceph-rbd_patch.yaml
+  - path: secretstores/secretstore-patch.yaml
--- a/cluster-scope/overlays/nerc-ocp-test/feature/odf/secretstores/secretstore-patch.yaml
+++ b/cluster-scope/overlays/nerc-ocp-test/feature/odf/secretstores/secretstore-patch.yaml
@@ -1,0 +1,10 @@
+apiVersion: external-secrets.io/v1beta1
+kind: SecretStore
+metadata:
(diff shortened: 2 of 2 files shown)


computate (approved):

computate (approved):

cluster-scope/overlays/nerc-ocp-test/feature/odf/components/nerc-secret-store/secretstore.yaml
larsks, line 1: To avoid confusion, I would avoid naming a directory "components" unless it actually contains Kustomize components (e.g., https://github.com/OCP-on-NERC/nerc-ocp-config/tree/main/cluster-scope/components).

larsks: Also, this should probably be consuming the `nerc-secret-store` component, and then just patch the `spec.provider.auth.kubernetes.mountPath` value, rather than replacing most of the resource:

```
patches:
  - patch: |
      apiVersion: external-secrets.io/v1beta1
      kind: SecretStore
      metadata:
        name: nerc-secret-store
      spec:
        provider:
          vault:
            auth:
              kubernetes:
                mountPath: kubernetes/nerc-ocp-test
```

(You could of course put this in a file instead of including it inline -- the salient change is that we're just patching the specific value that needs to be updated.)

larsks (changes requested):

cluster-scope/overlays/nerc-ocp-test/feature/odf/components/nerc-secret-store/secretstore.yaml
dystewart: @larsks In the interest of keeping the overlay as clean as possible added the streamlined patch via a file

larsks (approved):

PR #213, opened 2023-03-15
add custom routes for console and downloads

jtriley:
This is to restore the custom routes we had previously for the console and downloads routes in the `openshift-console` namespace.

### DIFF
--- a/cluster-scope/overlays/nerc-ocp-prod/feature/custom-routes/cluster-ingress.yaml
+++ b/cluster-scope/overlays/nerc-ocp-prod/feature/custom-routes/cluster-ingress.yaml
@@ -1,0 +1,13 @@
+apiVersion: config.openshift.io/v1
+kind: Ingress
+metadata:
+  name: cluster
+spec:
+  domain: apps.shift.nerc.mghpcc.org
+  componentRoutes:
+    - name: console
+      namespace: openshift-console
+      hostname: console.apps.shift.nerc.mghpcc.org
+    - name: downloads
+      namespace: openshift-console
+      hostname: downloads.apps.shift.nerc.mghpcc.org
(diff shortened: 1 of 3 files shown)


jtriley:
@larsks Do we want these manifests to update the existing/default routes in the `openshift-console` namespace? Also would we prefer to change this to `feature/console`?

computate (approved):

naved001 (approved):

larsks:
To change the routes for downloads and the console, you need to modify the Ingress configuration. See e.g. [this example](https://github.com/OCP-on-NERC/nerc-ocp-config/pull/200/files#diff-9527240e5e003f872ce11a14aeef1565b05a5b53be2a528933c6d7752e2b15b2R1) and [these docs](https://access.redhat.com/documentation/en-us/openshift_container_platform/4.8/html/web_console/customizing-web-console#customizing-the-console-route_customizing-web-console).

jtriley:
@larsks thanks for the pointers - should be fixed now. I also added the oauth route.

computate (approved):

cluster-scope/overlays/nerc-ocp-prod/feature/custom-routes/cluster-ingress.yaml
```diff
@@ -0,0 +1,16 @@
...
+spec:
+  domain: apps.shift.nerc.mghpcc.org
+  componentRoutes:
+    - name: console
+      namespace: openshift-console
+      hostname: console.apps.shift.nerc.mghpcc.org
+    - name: downloads
+      namespace: openshift-console
+      hostname: downloads.apps.shift.nerc.mghpcc.org
+    - name: oauth-openshift
```
larsks: Note that every time we change the oauth url, we also need to update any oauth application to which we're linked (e.g., Github, Keycloak), because the callback URL is part of the configuration.

I would probably just leave the oauth URL alone, since it's not something that we expect end users to care about...but I don't feel strongly about it.

jtriley: Fair point. Removed.

larsks (approved):

PR #214, opened 2023-03-15
Adding metrics and alerts for network transmit errors

computate:


cluster-scope/base/core/configmaps/thanos-ruler-custom-rules/configmap.yaml
```diff
@@ -72,6 +72,18 @@ data:
             PersistentVolumeClaim: "{{ $labels.persistentvolumeclaim }}"
             severity: warning
 
+        - alert: CustomNetworkInterfaceErrors
+          annotations:
+            summary: Node network transmit errors
+            description: "{{ $value }} node network transmit errors occured"
+          expr: node_network_transmit_errs_total > 0
```
aabaris: Sometimes an event occurs causing these counters to be non zero, but without an ongoing problem besides the event.  From there, network interface error counters persist until cleared via reboot.  

I would recommend to change this rule to something like:
expr: increase(node_network_transmit_errs_total[1h]) > 0

jtriley (approved):

naved001 (approved):

aabaris (changes requested):
Looks good.  I am making one request to adjust the alert rule (details in my comment for line 79 in configmap.yml) .  Please let me know if you have any questions or would like to discuss.

Thank you.

- Augustine

cluster-scope/base/core/configmaps/thanos-ruler-custom-rules/configmap.yaml
computate: Thanks for the insight @aabaris , I updated my commit with your suggestion!

aabaris:
Looks good.  Thank you!

aabaris (approved):

PR #216, opened 2023-03-16
Move ArgoCD "SkipDryRunOnMissingResource" patches to a component

larsks:
Back in bb8a93dc40 we attempted to move the "SkipDryRunOnMissingResource"
patches for ArgoCD into a separate file, but that didn't work because
(a) patches only effect resources generated by the `kustomization.yaml` in
which they are included, and (b) we had moved the patches into an isolated
subdirectory, so they were effectively a no-op.

But there is a better way! By making them a component [1] rather than a
resource, they are effectively included in the current kustomization.yaml.
This lets us both reduce the clutter in the nerc-ocp-prod overlay *and*
allows us to share the same set of patches in other overlays.

[1]: https://github.com/kubernetes/enhancements/blob/master/keps/sig-cli/1802-kustomize-components/README.md


larsks:
For verification, note that configuration generates output that is identical to its parent (8c97528ec01b920ade119bb10aca542ca96077a2).

dystewart (approved):
Good find! Feels much better without the clutter

naved001 (approved):

PR #217, opened 2023-03-16
fix NIC setup on nerc-ocp-test

jtriley:
This disables net.ifnames and biosdevnames in the kernel and adds udev rules for the network devices on the controllers and workers in nerc-ocp-test. This should result in consistent device names across the cluster which makes it easier to configure bond0 and other tagged interfaces (e.g. storage)


larsks (approved):

dystewart (approved):

jtriley:
Unfortunately I had to duplicate the bond0 machine config given that a role label is required (see https://github.com/openshift/machine-config-operator/issues/269). Another approach is to create a new machine config pool that includes both controllers + workers roles but decided against that for simplicity sake.

dystewart (approved):

PR #221, opened 2023-03-27
Adding new ipmi metrics to ACM Observability

computate:


larsks:
Do we really have to configure these statically? That is, is there no mechanism for either autodiscovery or some sort of pattern mechanism (e.g., include `ipmi_*`)?

naved001 (approved):

dystewart (approved):

PR #222, opened 2023-03-28
added root@nerc-admin01 key for coreos user

aabaris:


naved001 (approved):

cluster-scope/overlays/common/machineconfigs/99-master-ssh.yaml
```diff
@@ -17,3 +17,4 @@ spec:
...
+        - ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIG967CH5wXbesXHeLKgQFpMMLZ0l0QIjvkXNJvUlezwB nerc-admin01
```
larsks: The title says that this PR is adding keys for both @aabaris and @jtriley, but it's only adding a single key. Also, would you mind changing this entry so that the key comment is your username instead of `nerc-admin01`? Thanks!

larsks (changes requested):

cluster-scope/overlays/common/machineconfigs/99-master-ssh.yaml
aabaris: Originally I added mine and Justin's public keys, but then Justin suggested that adding key for root user for an admin host would be better.  I named it nerc-admin01 so not to reveal too much about where such a key would be.

I renamed the PR and the key, please let me know if this needs further clarification or changes.

jtriley (approved):

cluster-scope/overlays/common/machineconfigs/99-master-ssh.yaml
jtriley: The key belongs to root on nerc-admin01. This looks good now.

larsks: I generally recommend against shared keys, primarily because they make auditing harder (you lose information about who logged in when), and they make access control harder (when someone leaves we can't just remove their key; instead we need to generate a new shared private key).

PR #228, opened 2023-04-10
Adding ipmi and ceph custom metrics by regex wildcard

computate:
- I couldn't find documentation for this kind of regex matches, but I did find an [example that removes alerts matching a name and a job field](https://github.com/stolostron/multicluster-observability-operator/blob/3493964af4c4a583bdfff50740eb26322a5a1c95/examples/metrics/allowlist/custom-metrics-allowlist.yaml#L12-L13).
- I wondered if these kind of matches allow for the regular expression =~ matching, based on the [regex code I found here](https://github.com/stolostron/multicluster-observability-operator/blob/3493964af4c4a583bdfff50740eb26322a5a1c95/proxy/pkg/rewrite/rewrite.go#L23), and when I tried out the `matches` expression `- __name__=~"(ipmi_.*)"`, it worked!

### DIFF
--- a/cluster-scope/base/core/configmaps/observability-metrics-custom-allowlist/configmap.yaml
+++ b/cluster-scope/base/core/configmaps/observability-metrics-custom-allowlist/configmap.yaml
@@ -15,18 +15,6 @@ data:
       - node_network_transmit_bytes_total
       - node_network_transmit_errs_total
       - node_network_transmit_packets_total
-      - ipmi_bmc_info
-      - ipmi_chassis_power_state
-      - ipmi_current_amperes
-      - ipmi_current_state
-      - ipmi_dcmi_power_consumption_watts
-      - ipmi_fan_speed_rpm
-      - ipmi_fan_speed_state
-      - ipmi_power_state
-      - ipmi_power_watts
-      - ipmi_scrape_duration_seconds
-      - ipmi_sensor_state
-      - ipmi_sensor_value
-      - ipmi_temperature_celsius
-      - ipmi_temperature_state
-      - ipmi_up
+    matches:
+      - __name__=~"(ipmi_.*)"
+      - __name__=~"(ceph_.*)"


cluster-scope/base/core/configmaps/observability-metrics-custom-allowlist/configmap.yaml
```diff
@@ -15,18 +15,6 @@ data:
...
-      - ipmi_fan_speed_state
-      - ipmi_power_state
-      - ipmi_power_watts
-      - ipmi_scrape_duration_seconds
-      - ipmi_sensor_state
-      - ipmi_sensor_value
-      - ipmi_temperature_celsius
-      - ipmi_temperature_state
-      - ipmi_up
+    matches:
```
larsks, line 18: You mentioned that this syntax isn't documented. Do you know if it's *supported*? That is, can we rely on it working in the future?

larsks:
It would be great if you could update the commit message to include a link to the relevant section of code from which you derived this syntax.

cluster-scope/base/core/configmaps/observability-metrics-custom-allowlist/configmap.yaml
computate: Good call @larsks , as far as I can tell from the [similar example here](https://github.com/stolostron/multicluster-observability-operator/blob/3493964af4c4a583bdfff50740eb26322a5a1c95/examples/metrics/allowlist/custom-metrics-allowlist.yaml#L12-L13), although they don't use regex here, the `matches` expressions are supported:

computate:
Sure @larsks , I have updated with the commit with links from GitHub where I got the idea.

larsks:
@computate thanks for updating the commit message, but recall that commit messages need to render properly in the output of `git log` on a terminal, so using regular markdown formatting isn't usually a great idea.

I've reformatted the commit message as an example.

larsks (approved):

jtriley (approved):

PR #237, opened 2023-04-17
Adding RHODS image streams for project notebooks in test cluster

computate:
Here are some RHODS jupyter notebook images that the RHODS Red Hat Research team uses. 

- ucsls-stable-burosa-prods23 image
- stable-base-ope-nerc image

### DIFF
--- a/cluster-scope/overlays/nerc-ocp-test/kustomization.yaml
+++ b/cluster-scope/overlays/nerc-ocp-test/kustomization.yaml
@@ -13,3 +13,4 @@ resources:
 - machineconfigs/udev-rules
 - machineconfigs/configure-bond0
 - nodenetworkconfigurationpolicies/vlan-2175-nese.yaml
+- rhods/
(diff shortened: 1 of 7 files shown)


larsks (changes requested):
I would argue that these don't belong in the `cluster-scope` app, which is for cluster-wide configuration that requires admin privileges.

These ImageStream resources are namespaced resources specific to RHODS. There should probably be a top-level RHODS application, like we've done with [loki](https://github.com/OCP-on-NERC/nerc-ocp-config/tree/main/loki) or [vault](https://github.com/OCP-on-NERC/nerc-ocp-config/tree/main/vault), or a separate repository, like we did with [ipmi-metrics](https://github.com/OCP-on-NERC/ipmi-exporter).

dystewart:
@computate I agree with @larsks, for testing purposes I think we should create a top level application. Once we're offering RHODS in production it probably makes sense to make a dedicated repo.

computate:
Sounds good, I will close this PR for now, but it is an example of what to do for our separate application.

PR #241, opened 2023-05-02
Onboard ai4cloudops to test cluster

dystewart:
This commit adds the ai4cloudops project namespace to the test cluster, as well as adds a project-namespace-admin component that we can easily add to namespaces as needed moving forward. 
Syed is added as the project admin for now

### DIFF
--- a/cluster-scope/base/core/namespaces/ai4cloudops/kustomization.yaml
+++ b/cluster-scope/base/core/namespaces/ai4cloudops/kustomization.yaml
@@ -1,0 +1,6 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+- namespace.yaml
+- ../../../../components/project-admin-rolebindings/ai4cloudops
+namespace: ai4cloudops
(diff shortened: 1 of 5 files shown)


computate (approved):

dystewart:
@larsks confirming this is an active namespace, will need 1 more approval to get merged

larsks:
The commit message here doesn't really match the pr description.

cluster-scope/base/core/namespaces/ai4cloudops/kustomization.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+- namespace.yaml
+- ../../../../components/project-admin-rolebindings/ai4cloudops
```
larsks, line 5: I'm not sure about this: it's the *only* resource in `cluster-scope/base` that includes something from a parent directory. Do we really need a one-off here? We're not including these resources in any cluster overlays right now, but if this is intended for the production cluster we should be handling this via coldfront instead of by manually creating the namespace.

larsks (changes requested):

dystewart:
Closing as we're managing allocations and access with coldfront

PR #242, opened 2023-05-17
common: bump systemReserved to 4 CPU and 4Gi RAM

jtriley:
We're currently debugging an issue with timeouts/unavailability of at least the authentication cluster operator and potentially others (e.g. apiserver/kube-apiserver) on the nerc-ocp-infra cluster. In the case we opened with RedHat, they pointed out the warning about reserved system memory. This bumps the system memory from the default of 500m CPU and 1Gi RAM, via a kubeletconfig in the common overlay, to see if this helps given the number of pods running on this cluster.

See:
- https://github.com/OCP-on-NERC/operations/issues/127
- https://access.redhat.com/support/cases/#/case/03513513

### DIFF
--- a/cluster-scope/overlays/common/kubeletconfigs/kustomization.yaml
+++ b/cluster-scope/overlays/common/kubeletconfigs/kustomization.yaml
@@ -1,0 +1,2 @@
+resources:
+  - system-reserved.yaml
--- a/cluster-scope/overlays/common/kubeletconfigs/system-reserved.yaml
+++ b/cluster-scope/overlays/common/kubeletconfigs/system-reserved.yaml
@@ -1,0 +1,13 @@
+apiVersion: machineconfiguration.openshift.io/v1
+kind: KubeletConfig
+metadata:
+  name: system-reserved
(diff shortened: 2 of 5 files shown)


joachimweyl (approved):

larsks (approved):
Should we move this into the common overlay instead (so that it also applies to the prod cluster)?

jtriley:
> Should we move this into the common overlay instead (so that it also applies to the prod cluster)?

Sounds good - I've applied this manually for now on the infra cluster. Once we've cleared that I'll update this PR to move this into the common overlay.

jtriley:
@larsks This has been updated to use the common overlay and also apply to worker nodes on the prod cluster. I used a patch to avoid potentially needing to reboot controllers twice given that the infra cluster controllers are also part of the worker machine config pool. If you'd prefer to just add both controllers and workers in common instead let me know.

larsks:
> If you'd prefer to just add both controllers and workers in common instead let me know.

I guess I'm ambivalent about that. I think it's fine to go ahead and merge this as-is for now.

PR #243, opened 2023-05-18
Remove erroneous machineConfigPoolSelector configuration

larsks:
In pr #242, we added a Kubeletconfig resource with the following
configuration:

    machineConfigPoolSelector:
      matchLabels:
        pools.operator.machineconfiguration.openshift.io/master: ""
        pools.operator.machineconfiguration.openshift.io/worker: ""

That was intended to apply the configuration to both work and master nodes,
but from the documentation for the `machineConfigPoolSelector` key we see
that `matchLabels` are ANDed together:

    machineConfigPoolSelector    <Object>
      A label selector is a label query over a set of resources. The result
      of matchLabels and matchExpressions are ANDed. An empty label
      selector matches all objects. A null label selector matches no
      objects.

This commit *removes* the `machineConfigPoolSelector` configuration, which
means the Kubeletconfig will apply to all pools.

### DIFF
--- a/cluster-scope/overlays/nerc-ocp-prod/kubeletconfigs/system-reserved-patch.yaml
+++ b/cluster-scope/overlays/nerc-ocp-prod/kubeletconfigs/system-reserved-patch.yaml
@@ -5,6 +5,4 @@ metadata:
   namespace: openshift-config-operator
 spec:
   machineConfigPoolSelector:
-    matchLabels:
-      pools.operator.machineconfiguration.openshift.io/master: ""
-      pools.operator.machineconfiguration.openshift.io/worker: ""
+    $patch: delete


naved001 (approved):

jtriley (approved):

jtriley:
Just noting this will drain user-workloads on the production cluster. Ideally any deployments should fail-over but we don't necessarily know every user project is setup to be resilient to worker node reboots. Given that we should probably schedule a maintenance event before merge.

larsks:
> Just noting this will drain user-workloads on the production cluster. Ideally any deployments should fail-over but we don't necessarily know every user project is setup to be resilient to worker node reboots. Given that we should probably schedule a maintenance event before merge.

That's fine. I'll defer to you folks to schedule and communicate the downtime.

Should we schedule a regular (weekly?) maintenance window for this sort of minor change?

jtriley:
> That's fine. I'll defer to you folks to schedule and communicate the downtime.

Maintenance scheduled for May 31st to roll this out. https://github.com/nerc-project/nerc-docs/pull/112

> Should we schedule a regular (weekly?) maintenance window for this sort of minor change?

IMHO weekly is a bit too disruptive but that's something we can discuss at the next group meeting.

PR #244, opened 2023-05-18
Remove use of deprecated patchesStrategicMerge

larsks:
The Kustomize `patchesStrategicMerge` and `patchesJson6902` keywords
have both been replaced by the `patches` keyword.


larsks:
Sorry, erroneous pr.

PR #249, opened 2023-05-30
prod: fix system-reserved kubeletconfig patch

jtriley:
Not having a machineConfigPoolSelector defined in #243 results in the following error:

```
Error: could not find any MachineConfigPool set for KubeletConfig
```

In the spec docs it mentions that an empty label selector matches all objects:

> An empty label selector matches all objects. A null label selector matches no objects.

This patch attempts to set an empty label selector.


larsks (approved):
Worth a shot! I agree that the wording on those docs is a little vague; I was thinking we would need something that would yield:

```
machineConfigPoolSelector:
  matchLabels: []
```

Let's see what happens.

aabaris (approved):

PR #195, opened 2023-02-09
Add curator-operator and install prerequisites

dystewart:
This PR adds resources to deploy the curator operator, in this case specifically on the nerc-ocp-prod cluster. 

the first commit is adding the curator operator prerequisites. For this I've followed the instructions at [https://curator-operator.readthedocs.io/en/latest/requirements.html](https://curator-operator.readthedocs.io/en/latest/requirements.html)

PR #251, opened 2023-05-30
added dstewart changes to repo, fixed path issue in cluster-scope/bundles/curator-operator/kustomization.yaml

TRudenko22:
This is a resolution to pull request *Add curator-operator and install prerequisites #448*


larsks:
I'm updating #195 with these changes.

PR #256, opened 2023-06-30
Change cluster operator installPlanApproval to Manual to halt automatic upgrades

### LINKED ISSUE #148: Update cluster operators to require manual updates
We want to make sure updates happen when people are available to watch them

@computate @aabaris and I ran into a pod from the lokistack which was suddenly broken. This break caused memory resources of one of the control plane nodes to be completely consumed which had varying negative effects on infra cluster performance. 

We traced the issue back to one of the associated operators having upgraded automatically in the background. We should change the installPlanApproval field to manual for our operators to prevent something like this from happening in the future. 

In order to manually update operators we can follow this procedure [https://docs.openshift.com/container-platform/4.10/operators/admin/olm-upgrading-operators.html#olm-approving-pending-upgrade_olm-upgrading-operators](https://docs.openshift.com/container-platform/4.10/operators/admin/olm-upgrading-operators.html#olm-approving-pending-upgrade_olm-upgrading-operators)

dystewart:
With  manual updates, when a newer version of an Operator is available, OLM creates an update request. As a cluster administrator, you must then manually approve that update request to have the Operator updated to the new version.

Closes: [https://github.com/OCP-on-NERC/operations/issues/148](https://github.com/OCP-on-NERC/operations/issues/148)

### DIFF
--- a/cluster-scope/base/operators.coreos.com/subscriptions/acm/subscription.yaml
+++ b/cluster-scope/base/operators.coreos.com/subscriptions/acm/subscription.yaml
@@ -4,7 +4,7 @@ metadata:
     name: acm
 spec:
     channel: release-2.6
-    installPlanApproval: Automatic
+    installPlanApproval: Manual
     name: advanced-cluster-management
     source: redhat-operators
     sourceNamespace: openshift-marketplace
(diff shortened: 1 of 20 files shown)


larsks:
Please make sure that commit message is *at least* as descriptive as the PR description :)

larsks:
I'm not opposed to this, and the PR looks fine syntactically, but it represents a substantial change in how we manage the cluster, and in particular means increased effort on the part of the people maintaining the cluster.

While I think major upgrades should be planned in advance (like a cluster upgrade from 4.x -> 4.(x+1)), I think we ought to be able to automatically apply upgrades as long as we're pinned to either a "stable" channel or a versioned channel (e.g., our subscription for the gitops operator is for the `gitops-1.6` channel, so all updates should be in with 1.6.x series).

dystewart:
Closing this based on feedback from the team and not wanting to add additional overhead to managing the clusters

PR #261, opened 2023-07-12
nerc-ocp-test: add gpu operator and clusterpolicy

jtriley:
This installs the nvidia-gpu-operator bundle and a ClusterPolicy resource on the nerc-ocp-test cluster for testing out GPU support on OpenShift.


aabaris (approved):

jtriley:
Just noting I have included the clusterpolicy in the overlay, however, as far as I know, the operator isn't really usable without a policy. The clusterpolicy included in this PR is the default. We might want to move that into the base bundle and then overlays can customize as needed via patching. Thoughts? cc @larsks

dystewart (approved):

larsks:
@jtriley I think it's fine where it is right now, but if we find ourselves needing to deploy this elsewhere we'll probably want to move it.

larsks (approved):

PR #263, opened 2023-07-14
Add curator resources to repo and deploy to nerc-ocp-test through test cluster overlays

dystewart:
These are all the resources specific to running the curator deployment. There are a few and it's a rather lengthy list of items so I've tried to split this up into more manageable commits

All in all we're:

1. Adding and configuring curator-system namespace
2. Adding curator crds 
3. Adding curator specific namespace-scoped resources
4. Adding and configuring curator postgres database deployment
5.  Adding curator resources to the nerc-ocp-test overlays


curator/base/deploymentconfigs/deploymentconfig.yaml
```diff
@@ -0,0 +1,54 @@
+apiVersion: apps.openshift.io/v1
+kind: DeploymentConfig
```
larsks: I would prefer to see Deployment resources, which are native to Kubernetes, rather than OpenShift-specific DeploymentConfig resources. Red Hat makes the same recommendation in [their docs on the topic](https://docs.openshift.com/container-platform/4.10/applications/deployments/what-deployments-are.html) ("it is recommended to use Deployment objects unless you need a specific feature or behavior provided by DeploymentConfig objects.").

larsks (changes requested):

curator/base/deploymentconfigs/deploymentconfig.yaml
larsks: It would also be nice to see a backup job for the curator data, but that can be part of a future pull request.

larsks:
@dystewart you just undid all the commit changes we did earlier in the week :(. I think you forgot to update your local repository before making additional changes.

larsks:
@dystewart it looks like there's a merge here, which shouldn't have happened, and the "fix kustomize error" would ideally be folded into one of the existing commits (like we did earlier in the week).

Fixing things:

```
# reset history to last commit from Thursday
git reset d23eb7b6e115515d3edea10711ae1e82f3da3056

# this leaves us with one modified file. add it...
git add curator/base/kustomization.yaml

# and make a --fixup commit (this looks for the first
# commit that contains "Add top-level curator")
git commit --fixup ":/Add top-level curator"

# rebase on main branch (this may be something other
# than "upstream/main" depending on how your local
# repository is configured). This should have
# automatically placed the --fixup commit in the correct
# place and marked it as "fixup". Just save the
# pick list without making changes.
git rebase -i upstream/main
```

I have pushed these changes to the PR.

larsks:
Before making any additional changes, you will want to:

1. `git remote update`
1. `git checkout curator-operator`
1. `git reset --hard origin/curator-operator`

(Assuming that `origin` points to your github remote)

This ensures that your local `curator-operator` branch matches the content of this PR.

dystewart:
Thanks for all the help @larsks, I haven't really gotten down into the weeds of fixing up a PR like this so I really appreciate the help. Everything should be good to go at this point

larsks:
@dystewart after creating a `fixup!` commit you need to `git rebase -i` and merge it into the appropriate existing commit, rather than making it part of the pull request.

dystewart:
Aha fixed @larsks

larsks (approved):

jtriley (approved):

PR #264, opened 2023-07-14
Add koku-metrics operator manifests and add operator to nerc-ocp-test overlays

dystewart:
Adding koku-metrics operator subscription to test cluster to work with the curator deployement

### DIFF
--- a/koku-metrics-operator/base/kokumetricscfg.yaml
+++ b/koku-metrics-operator/base/kokumetricscfg.yaml
@@ -1,0 +1,14 @@
+apiVersion: koku-metrics-cfg.openshift.io/v1beta1
+kind: KokuMetricsConfig
+metadata:
+  name: kokumetricscfg
+spec:
+  authentication: {}
+  packaging:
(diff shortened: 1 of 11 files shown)


koku-metrics-operator/base/kokumetricscfg.yaml
jtriley, line 1: Do we want to move this to the cluster-scope overlay instead? Otherwise we would need to add this to the app-of-apps repo, no?

jtriley (approved):

larsks (approved):

koku-metrics-operator/base/kokumetricscfg.yaml
larsks: Sorry, I merged this not realizing you had a question here. 

I think that doing this **not** in the cluster-scope overlay is probably the right decision (it prevents cluster-scope from becoming a dumping ground). It will need to be added to the apps repo, and I'm happy to revert the merge if necessary.

PR #265, opened 2023-07-21
Remove cluster specific config from secretstore component

### LINKED ISSUE #130: Secretstore cluster-scope component points to kubernetes/nerc-ocp-prod by default
Now that we have secretstores in more than one cluster (both prod and test) we should change this [line](https://github.com/OCP-on-NERC/nerc-ocp-config/blob/879cf90e69dc074274d098878661bd96c382edf3/cluster-scope/components/nerc-secret-store/secretstore.yaml#L13) and specify that the secretstore should be patched in overlays.

larsks:
Remove the vault auth mountPath for nerc-ocp-prod from the secretstore component.
This will result in an invalid configuration unless the value is changed in
an overlay.

Closes ocp-on-nerc/operations#130

### DIFF
--- a/cluster-scope/components/nerc-secret-store/secretstore.yaml
+++ b/cluster-scope/components/nerc-secret-store/secretstore.yaml
@@ -10,7 +10,7 @@ spec:
       version: "v2"
       auth:
         kubernetes:
-          mountPath: kubernetes/nerc-ocp-prod
+          mountPath: UPDATE IN OVERLAY
           role: secret-reader
           secretRef:
             name: "vault-secret-reader"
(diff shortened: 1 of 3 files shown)


larsks:
@dystewart note that this will **change** the configuration on nerc-ocp-test, since that cluster was using an incorrect value.

larsks:
@dystewart note that this will **change** the configuration for nerc-ocp-test, since that cluster was previously using an incorrect value:

```diff
--- cluster-scope/overlays/nerc-ocp-test/before.yaml    2023-07-21 10:41:54.474478943 -0400
+++ cluster-scope/overlays/nerc-ocp-test/after.yaml     2023-07-21 10:42:06.456536564 -0400
@@ -590,7 +590,7 @@
     vault:
       auth:
         kubernetes:
-          mountPath: kubernetes/nerc-ocp-prod
+          mountPath: kubernetes/nerc-ocp-test
           role: secret-reader
           secretRef:
             key: token
@@ -611,7 +611,7 @@
     vault:
       auth:
         kubernetes:
-          mountPath: kubernetes/nerc-ocp-prod
+          mountPath: kubernetes/nerc-ocp-test
           role: secret-reader
           secretRef:
             key: token
```

dystewart (approved):

naved001 (approved):

PR #266, opened 2023-07-28
Add udev rules for gpu nodes

larsks:
We configure the bond0 interface to span interfaces named "nic1" and
"nic2", and we rely on udev rules to assign these names to the appropriate
interfaces. There were no such rules for the GPU nodes (wrk-{10,11}) on the
test cluster.

This commit adds the necessary udev rules.

Part of: ocp-on-nerc/operations#170


jtriley (approved):

dystewart (approved):

cluster-scope/overlays/nerc-ocp-test/machineconfigs/udev-rules/src/bcm57412.rules
```diff
@@ -0,0 +1,2 @@
+SUBSYSTEM=="net",DRIVERS=="bnxt_en",KERNELS=="0000:18:00.0",ACTION=="add",NAME="nic1"
+SUBSYSTEM=="net",DRIVERS=="bnxt_en",KERNELS=="0000:b3:00.1",ACTION=="add",NAME="nic2"
```
larsks: I need to check the value of `KERNELS` on the second line to make sure that difference in the second position isn't a copy/paste error.

larsks: Fixed; that was a typo.

PR #279, opened 2023-09-13
Move dex configuration to .spec.sso.dex

larsks:
The gitops operator was complaining about our argocd manifest:

    Illegal expression of SSO configuration detetected for Argo CD
    openshift-gitops in namespace openshift-gitops. cannot specify
    .spec.Dex fields when dex is configured through .spec.sso.dex

This commit moves the Dex configuration from .spec.dex to .spec.sso.dex to
try to resolve the error.


larsks:
I'm going to resubmit this...

PR #284, opened 2023-09-21
prod: fix NIC device names via udev rules

jtriley:
Previously we had homogeneous worker node hardware, however, with the addition of 2x new NVIDIA V100 worker hosts this is no longer the case. These new hosts have different NIC devices than the rest of the worker nodes.

This adds the following new machineconfigs in order to map all of the NICs in the cluster to nic1 and nic2 devices:

- disable-net-ifnames
- custom-udev-rules

The configure-bond0 machineconfig has been updated to use nic1 and nic2 vs the old eno1 and eno2 device names. Previously, this machineconfig was only being applied on the worker nodes. This patch also extends that configuration to the controllers in order to setup an LACP bond on those hosts.


cluster-scope/overlays/nerc-ocp-prod/machineconfigs/configure-bond0/bond0-controller.yaml
```diff
@@ -0,0 +1,28 @@
...
+  name: configure-bond0-controller
+spec:
+  config:
+    ignition:
+      version: 3.2.0
+    storage:
+      files:
+        - contents:
+            compression: gzip
+            source: data:;base64,H4sIAAAAAAAC/0zMwaqDMBCF4f15FyXeKxfvYp5EskiTKQY6M6IToW9fLEK7O/BzvjmbKmevphG10M20BPhz5fdEVeftnjJ3moSvnJrb9SPfGuOD9F+t2x/p4J0GYGZfeFP2CPFG/yEEYD61CLHCNIWf/jcVSK1iSsN49roeY4SwL1bohFGWvHZeha05TX9jCHgFAAD///WlE5PCAAAA
```
aabaris, line 16: @jtriley how come bond0 file is encoded/compressed, but nic1 and nic2 are not?

computate (approved):

cluster-scope/overlays/nerc-ocp-prod/machineconfigs/configure-bond0/bond0-controller.yaml
jtriley: Good question - this file gets generated by butane in the `Makefile`. My guess is that is governed by the size of the result. According to line count, bond0 file is larger than nic1 and nic2:

```
$ wc -l src/*.nmconnection
      17 src/bond0.nmconnection
      10 src/nic1.nmconnection
      10 src/nic2.nmconnection
```

aabaris (approved):

PR #314, opened 2023-11-07
Add openapi schema for proper patching of CRDs

larsks:
Due to issues with the way Kustomize handles patching lists in custom
resource types, we need to provide a complete OpenAPI schema that
includes schema definitions for any CRDs we wish to patch.

See `cluster-scope/base/openapi/README.md` for more information.

### DIFF
--- a/cluster-scope/base/openapi/README.md
+++ b/cluster-scope/base/openapi/README.md
@@ -1,0 +1,97 @@
+# Why is this OpenAPI schema here?
+
+Many kubernetes resources include lists of objects, such the
+`spec.containers` field of Pods:
+
+    spec:
+      containers:
+        - name: container1
+          image: example
+        - name: container2
+          image: example
+
+Kustomize is able to apply a strategic merge patch to specific
+elements in this list by matching the patch against a "merge key".
+That is, if we want to add an `env` attribute to `container2`, we can
+create a patch like this:
+
+    apiVersion: v1
+    kind: Pod
+    metadata:
+      name: example
+    spec:
+      containers:
+        - name: container2
+          env:
+            - MYVAR=somevalue
+
+Kustomize "knows" to patch the list item with `name` equal to
+`container2` because it has a compiled in schema for native kubernetes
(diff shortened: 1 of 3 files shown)


naved001 (approved):

computate:
@larsks is the openshift-api-schema.json file specific to a certain cluster? What if we want to patch a GrafanaDataSource CRD on the upcoming observability cluster that doesn't manage secrets well and requires the Patch Operator: 
```yaml
apiVersion: integreatly.org/v1alpha1
kind: GrafanaDataSource
```

larsks:
> is the openshift-api-schema.json file specific to a certain cluster? 

It is not. It needs to be a superset of the resources available in all of our clusters (at least those we want to patch).

> What if we want to patch a GrafanaDataSource CRD on the upcoming observability cluster that doesn't manage secrets well and requires the Patch Operator:

Then we would need to add the appropriate schemas to the file.

---

I've partly addressed this in the README:

> To support this, that means we need to dump the complete openapi schema
from our clusters to a file and make that available to Kustomize. This
commit introduces an initial version of this schema, but we will need
to update this in the future whenever we want to patch CRDs that
weren't available at the time we last generated this schema dump.

larsks:
I've had a change of heart about how to implement this; I've moved all the sausage making to <https://github.com/nerc-project/openshift-schemas>; the generated schema is included here by URL.

dystewart (approved):

.pre-commit-config.yaml
```diff
@@ -12,6 +12,7 @@ repos:
       - id: check-merge-conflict
       - id: end-of-file-fixer
       - id: check-added-large-files
+        exclude: ^cluster-scope/base/openapi/
```
jtriley: Do we still need this given that the schema json now lives in a separate repository?

larsks: Good catch; we don't need this anymore. I'll take it out.

computate:
That new repo is very clever, I like your sausage recipe.

PR #316, opened 2023-11-09
Refactor oauth configuration

larsks:
We're using a similar oauth configuration on all of our clusters (typically
consisting of one or both of GitHub and the NERC Keycloak instance). This
commits factors out the oauth configuration into a set of common components
so that instead of replicating the same configuration in multiple overlays,
we can include one or more reusable components.

E.g., for a cluster that only needs keycloak authentication:

    components:
      - ../../components/nerc-oauth-keycloak

Or for a cluster that wants both keycloak and GitHub:

    components:
      - ../../components/nerc-oauth-keycloak
      - ../../components/nerc-oauth-github

In either case, the overlay would apply the necessary patches to override
things like client ids, secret names, etc.


knikolla (approved):
I like this, thanks Lars!

naved001 (approved):

dystewart (approved):

schwesig (commented):
/lgtm
/approve

PR #317, opened 2023-11-13
Custom grafana dashboard

jbasu01:
A new custom Grafana dashboard was created and tested by bringing in metrics from ArgoCD. ArgoCD provides a JSON template download, and code from the JSON download could be copied into the config file and appropriate changes made to ensure metrics from ArgoCD dashboard are pulled in and made visible in the custom Grafana dashboard. Additional information is available at https://grafana.com/grafana/dashboards/14584-argocd/

A key item to make a note of is that one cannot use the word "ArgoCD" or "argocd" just by itself in the value for the key "title" in the grafana-dashboard-argocd.yaml.  When changed to "ArgoCD xxxxx" ("ArgoCD stats" in our case and it worked fine.


schwesig:
@schwesig

larsks:
> A key item to make a note of is that one cannot use the word "ArgoCD" or "argocd" just by itself in the value for the key "title" in the grafana-dashboard-argocd.yaml.

Why?

schwesig (approved):
working together with Jeet in a call
checking and linting together

grafana/base/grafanadatasources/grafanadatasource.yaml
```diff
@@ -0,0 +1,38 @@
+apiVersion: integreatly.org/v1alpha1
+kind: GrafanaDataSource
+metadata:
+  name: loki-datasource
+  namespace: grafana
+spec:
+  name: loki
+  datasources:
+    - name: application-logging
```
computate: We need to remove the Loki datasources and replace it with the prometheus datasource instead, and make sure the Patch Operator is working to patch the Bearer token in here: 
```yaml
  datasources:
    - access: proxy
      editable: false
      isDefault: true
      jsonData:
        httpHeaderName1: Authorization
        timeInterval: 5s
        tlsSkipVerify: true
      name: Prometheus
      secureJsonData:
        httpHeaderValue1: >-
          Bearer ...
      type: prometheus
      url: 'https://thanos-querier.openshift-monitoring.svc.cluster.local:9091/'
  name: custom-observability-grafana
  ```

**Comment by computate** [comment]
@jbasu01 @schwesig Nice job resolving the pre-commit hooks and squashing to one commit. I just want to test this out together in an upcoming meeting on Monday, update the datasources, and make sure it's ready. I realize @jbasu01 will be on vacation, but we can work out the final details next week.

**Comment by computate** [comment]
@larsks after a couple hours going crazy testing out a custom dashboard in the ACM grafana, I can't explain why, but naming the dashboard anything but `argocd` or `ArgoCD` magically made the dashboard work.

**Comment by schwesig** [comment]
@computate thanks for checking the argocd naming problem we had and confirming it.

File: grafana/base/grafanadatasources/grafanadatasource.yaml [hunk 2]
```diff
@@ -0,0 +1,38 @@
...
+        httpHeaderName1: X-Scope-OrgID
+      secureJsonData:
+        httpHeaderValue1: application
+    - name: infrastructure-logging
+      type: loki
+      url: >-
+        https://lokistack-querier-http.openshift-operators-redhat.svc.cluster.local:3100
+      withCredentials: true
+      jsonData:
+        tlsSkipVerify: true
```
larsks: Are you sure this is necessary? You should be able to use the ca cert provided in `/run/secrets/kubernetes.io/serviceaccount` to trust other services on the cluster without disabling tls verification.

grafana/base/grafanas/grafana.yaml
```diff
@@ -0,0 +1,26 @@
...
+  name: grafana
+  namespace: grafana
+spec:
+  deployment:
+    envFrom:
+      - configMapRef:
+          name: grafana-config-overrides
+  config:
+    server:
+      root_url: $GRAFANA_ROOT_URL
```
larsks: I would avoid placeholders that look like variables (because some might be confused about how things operate). Consider something like...

```suggestion
      root_url: REPLACE_IN_OVERLAY
```

grafana/base/kustomization.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+  - operatorgroups
+  - subscriptions
+  - configmaps
+  - grafanas
+  #  - grafanadatasources
```
larsks: If we don't need this, remove it. If it's meant as an example, then add a comment documenting why it's here.

grafana/base/routes/route.yaml
```diff
@@ -0,0 +1,17 @@
+kind: Route
+apiVersion: route.openshift.io/v1
+metadata:
+  name: grafana
+  namespace: grafana
+spec:
+  host: grafana.apps.nerc-ocp-infra.rc.fas.harvard.edu
```
larsks: You should treat this the same way you treat `root_url` in `base/grafanas/grafana.yaml`.

grafana/base/serviceaccounts/serviceaccount.yaml
```diff
@@ -0,0 +1,6 @@
+---
```
larsks: Remove the YAML document separator (you're not using it anywhere else).

grafana/overlays/nerc-ocp-infra/kustomization.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+resources:
+  - ../../base
+  - externalsecrets
+  - patches/logging-grafana-patch.yaml
+
+patchesStrategicMerge:
+  - grafanas/logging-grafana_patch.yaml
```
larsks: This should be:

```suggestion
patches:
- path: grafanas/logging-grafana_patch.yaml
```

`patchesStrategicMerge` has been [deprecated for a while](https://github.com/kubernetes-sigs/kustomize/issues/5052).

larsks (changes requested):

schwesig:
@larsks Thanks for your feedback.
@jbasu01 next steps to work on for us

grafana/base/grafanas/grafana.yaml
computate: I have applied this suggestion.

grafana/base/routes/route.yaml
computate: I have applied this suggestion.

grafana/base/serviceaccounts/serviceaccount.yaml
computate: I have applied this suggestion.

grafana/base/kustomization.yaml
computate: I have applied this suggestion.

computate:
@larsks I'm having trouble with the Patch Operator with this PR. I have applied the changes from this branch to test it: 

```
oc --as system:admin apply -k grafana/overlays/nerc-ocp-infra/
```

The Patches say `LastReconcileCycleSucceded`: 

```bash
oc --as system:admin -n grafana get patch/grafana-oauth -o yaml | grep 'status:' -A 20
oc --as system:admin -n grafana get patch/grafanadatasource-observability-metrics -o yaml | grep 'status:' -A 20
```

But the objects are not patched: 

```bash
oc --as system:admin -n grafana get grafana/grafana -o yaml | grep 'root_url:'
oc --as system:admin -n grafana get grafanadatasource/observability-metrics -o yaml | grep 'secureJsonData:' -A 2
```

@larsks Can you please help me understand why the patches are not working, and if I need this ClusterRoleBinding? 

```yaml
kind: ClusterRoleBinding
metadata:
  name: grafana-node-labeler-patcher
roleRef:
  kind: ClusterRole
  name: node-labeler
subjects:
  - kind: ServiceAccount
    name: patcher
    namespace: grafana
```

larsks:
@computate, looking at the logs for the patch operator, we see the following errors after creating the Patch resource:

```
...: failed to list /v1, Kind=Secret: secrets is forbidden: User
"system:serviceaccount:grafana:patcher" cannot list resource "secrets" in API
group "" at the cluster scope

...: failed to list integreatly.org/v1alpha1, Kind=Grafana:
grafanas.integreatly.org is forbidden: User
"system:serviceaccount:grafana:patcher" cannot list resource "grafanas" in API
group "integreatly.org" at the cluster scope

...: Failed to watch /v1, Kind=Secret: failed to list /v1, Kind=Secret: secrets
is forbidden: User "system:serviceaccount:grafana:patcher" cannot list resource
"secrets" in API group "" at the cluster scope

...: Failed to watch integreatly.org/v1alpha1, Kind=Grafana: failed to list
integreatly.org/v1alpha1, Kind=Grafana: grafanas.integreatly.org is forbidden:
User "system:serviceaccount:grafana:patcher" cannot list resource "grafanas" in
API group "integreatly.org" at the cluster scope
```

It looks like a permissions issue; we'll need to add some RBAC to provide the necessary permissions.

...but the `node-labeler` role isn't appropriate, because you're not labelling nodes.

larsks:
Add this RBAC to your pr and things seem to work as expected:

```
kind: ClusterRole
metadata:
  name: grafana-patcher
rules:
- apiGroups:
  - ""
  resources:
  - secrets
  verbs:
  - get
  - list
  - watch
  - update
  - patch
- apiGroups:
  - integreatly.org
  resources:
  - grafanas
  verbs:
  - get
  - list
  - watch
  - update
  - patch
---
kind: ClusterRoleBinding
metadata:
  name: grafana-patcher
roleRef:
  kind: ClusterRole
  name: grafana-patcher
subjects:
- kind: ServiceAccount
  name: patcher
  namespace: grafana
```

Following our organizational model, that should go into the `cluster-scope` application somewhere.

larsks:
> Following our organizational model, that should go into the cluster-scope application somewhere.

...or not, since I see the `grafana` application includes the subscription and other resources we would normally put in `cluster-scope`, so it's best to keep things together.

larsks:
@computate, looking at <https://github.com/grafana-operator/grafana-operator/blob/v3.6.0/documentation/env_vars.md> this patching may no longer be necessary. I'm not 100% positive; followed <https://github.com/grafana-operator/grafana-operator/issues/130> to find that link and I'm taking a closer look now.

larsks:
I think you ought to be able to do something like this...

In the Grafana resource:

```
  deployment:
    envFrom:
    - configMapRef:
        name: grafana-config-overrides
    - secretRef:
        name: grafana-secrets
```

And in grafana-secrets:

```
apiVersion: v1
kind: Secret
metadata:
  name: grafana-secrets
type: Opaque
data:
  GF_AUTH_GENERIC_OAUTH_CLIENT_SECRET: "..."
  GF_AUTH_GENERIC_OAUTH_CLIENT_ID: "..."
```

You should be patching `spec.config.server` via Kustomize, not via a dynamic Patch resource.

larsks:
@computate with this latest set of changes I think everything is working (and no more patch operator required).

computate (approved):
@larsks Nice work with the secretGenerator, ExternalSecrets, and environment variable substitutions, thanks!

schwesig (approved):
@larsks awesome work, thanks for having these ideas/digging them up. appreciated.
Also thanks for getting me all the details in our call.

grafana/overlays/nerc-ocp-infra/grafanadatasources/observability-metrics.yaml
```diff
@@ -0,0 +1,21 @@
...
+      isDefault: true
+      jsonData:
+        httpHeaderName1: Authorization
+        timeInterval: 5s
+        tlsAuthWithCACert: true
+      secureJsonData:
+        httpHeaderValue1: "Bearer ${token}"
+        tlsCACert: "${service-ca.crt}"
+      type: prometheus
+      url: 'https://thanos-querier.openshift-monitoring.svc.cluster.local:9091/'
```
computate: @harshil-codes pointed out that we should change the url here to be the correct observability thanos querier `http://observability-thanos-query.open-cluster-management-observability.svc.cluster.local:9090`

computate: @schwesig @larsks  I just added a commit to change this URL while making another change to add a ClusterRoleBinding for the grafana-servicaccount to read the metrics.

schwesig (approved):

schwesig (approved):

larsks (approved):

PR #318, opened 2023-11-14
Define APIServer resource in base and patch in overlays

larsks:
Our overlays were all performing effectively the same changes to the
APIServer resource in order to enable a custom certificate. This
commit moves APIServer into our base manifests and adds appropriate
per-overlay patches to update the hostname.


aabaris (approved):

schwesig (approved):
/lgtm
/approve

PR #328, opened 2023-11-27
Standardize our handling of custom certificates

larsks:
We were handling custom ingress and apiserver certificates in different ways, and we were also applying identical patches in multiple overlays.

This commit re-organizes custom API and Ingress certificates into a single "custom-certificates" component, and arranges to include that in our common overlay.


naved001 (approved):

knikolla (approved):
Looks good to me. `httpEmptyRequestsPolicy: Respond` and 
`replicas: 2` seem to be defaults of the Ingress Operator.

PR #330, opened 2023-11-27
Ensure that we deploy a MetalLB resource

larsks:
After installing the MetalLB operator, we must create a "MetalLB" resource instance. See [1] for documentation.

[1]: https://docs.openshift.com/container-platform/4.14/networking/metallb/about-metallb.html


schwesig (approved):
checked against [1](https://docs.openshift.com/container-platform/4.14/networking/metallb/about-metallb.html), looking good

dystewart (approved):

PR #331, opened 2023-11-27
Consistently manage admin acks

larsks:
We eventually require the admin-acks ConfigMap [1] in all of our clusters. Rather than duplicating it everywhere, including it in base and use a configMapGenerator to populate it in each overlay.

[1]: https://docs.openshift.com/container-platform/4.14/updating/preparing_for_updates/updating-cluster-prepare.html#update-preparing-ack_updating-cluster-prepare


schwesig (approved):
checked changes, looks good to me

aabaris (approved):

PR #332, opened 2023-11-28
Remove deprecated syntax

larsks:
The `patchesStrategicMerge` keyword has been deprecated for a while [1]; we
should be using the `patches` keyword for both strategic merge and
JSONPatch patches.

[1]: https://github.com/kubernetes-sigs/kustomize/issues/5052


schwesig (approved):
checked, looking good to me

computate (approved):

PR #338, opened 2024-01-09
Enable rhods notebook controller in rhods-notebooks namespace

dystewart:
This will re-enable the default functionality of the rhods notebook controller in the prod cluster. As a side effect of enabling, the ability to launch a notebook from the jupyter tile will be restored 
Addresses: https://github.com/nerc-project/operations/issues/361


DanNiESh (approved):
LGTM

computate (approved):

PR #341, opened 2024-01-15
Add ope-course-quota resourceQuota to cluster-scope components

### LINKED ISSUE #377: Create resourceQuota in rhods-notebooks namespace (Prod cluster)
We need to create a resourceQuota in this namespace since there isn't one by default. 

We will need a pretty hefty quota since both ope courses are using the namespace. Since we can't limit the notebook resource options to the minimal allocation in rhods without removing them as options from all the users, we need to keep in mind that students technically have the ability to choose any of the allocation options for their notebook. 

We have in the [docs](https://massopencloud.slack.com/files/U04QKBPMUQ6/F06DJUR7GUA/how_to_launch_ucsls_jupyter_instance.pdf?origin_team=T14CVG7L1&origin_channel=CNMSRMDA5) specified the x-small resource option but we can't enforce this in our current setup. That makes it somewhat tricky to nail down an exact hard quota limit Since in the absolute worst case (incredibly unlikely but technically possible) all students could select other bigger options. 

 We ran the scale test in ope-rhods-testing with 300 instances and didn't hit our quota so maybe that's a good starting point. As for storage we can pretty comfortably set a limit once we know exactly the number of students we expect (all notebooks create a pvc with one Gi). 

We can use this as a starting point:
```
kind: ResourceQuota
apiVersion: v1
metadata:
  name: spring-course-quota
  namespace: rhods-notebooks

spec:
  hard:
    limits.cpu: '1000'
    limits.ephemeral-storage: 50Gi
    limits.memory: 3000000Mi
    persistentvolumeclaims: '1000'
    requests.storage: 100Gi
```

We should run the test again with the expected number of students to see how it looks against something like the resourceQuota above.

dystewart:
Closes: [https://github.com/nerc-project/operations/issues/377](https://github.com/nerc-project/operations/issues/377)

Adding this resourceQuota as a component at the cluster scope and adding it to the rhods-notebooks namespace in the prod cluster. The quota values themselves were determined based on our resourceQuota in the namespace where we did scale testing, with a bump to accomodate all the expected students

### DIFF
--- a/cluster-scope/components/resourcequotas/ope-course-quota/kustomization.yaml
+++ b/cluster-scope/components/resourcequotas/ope-course-quota/kustomization.yaml
@@ -1,0 +1,4 @@
+apiVersion: kustomize.config.k8s.io/v1alpha1
+kind: Component
+resources:
+- resourcequota.yaml
--- a/cluster-scope/components/resourcequotas/ope-course-quota/resourcequota.yaml
+++ b/cluster-scope/components/resourcequotas/ope-course-quota/resourcequota.yaml
@@ -1,0 +1,12 @@
+kind: ResourceQuota
+apiVersion: v1
+metadata:
+  name: ope-course-quota
+  namespace: rhods-notebooks
+spec:
+  hard:
+    limits.cpu: '1000'
+    limits.ephemeral-storage: 30Gi
+    limits.memory: 3000000Mi
+    persistentvolumeclaims: '400'
+    requests.storage: 400Gi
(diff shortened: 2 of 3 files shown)


computate (approved):
Clever to add a resource quota to prevent large allocations.

dystewart:
Most of our resourceQuotas are created via ColdFront so this is a bit of a one-off but I think it makes the most sense to add this as a cluster-scope component

IsaiahStapleton (approved):
Looks good

hpdempsey:
Looks good.  Also something @schwesig should take into account for the class monitoring and alerting events, so we know when we are "approaching" a quota problem.

schwesig:
> Looks good.  Also something @schwesig should take into account for the class monitoring and alerting events, so we know when we are "approaching" a quota problem.
> 
> 

ack

schwesig (approved):
Follows discussion in slack, makes sense👍

PR #342, opened 2024-01-17
Configure test cluster to use LOKI from observability cluster

### LINKED ISSUE #324: Setup Logging on the New Observability Cluster 2/2
- subtask of https://github.com/nerc-project/operations/issues/177
- Zenhub Epic [```Setup a new Cluster to handle Metrics, Logging and Observability```](https://app.zenhub.com/workspaces/moc-a--rh-sprint-workspace-62a210f69d42f600151deae0/epics/Z2lkOi8vcmFwdG9yL1plbmh1YkVwaWMvMTQzNzMy)

timeline: 2023/Q4 - 2024/Q1

right now catching this as an issue for #177

- [x] logging solution in #405
- [ ] log sources this issue

bnshr:
fixes https://github.com/nerc-project/operations/issues/324
Referring to https://github.com/nerc-project/operations/issues/324

### DIFF
--- a/cluster-scope/base/user.openshift.io/groups/cs210/group.yaml
+++ b/cluster-scope/base/user.openshift.io/groups/cs210/group.yaml
@@ -1,5 +1,0 @@
-apiVersion: user.openshift.io/v1
-kind: Group
-metadata:
-  name: cs210
-users: []
--- a/cluster-scope/base/user.openshift.io/groups/ee440/group.yaml
+++ b/cluster-scope/base/user.openshift.io/groups/ee440/group.yaml
@@ -1,5 +1,0 @@
-apiVersion: user.openshift.io/v1
-kind: Group
-metadata:
-  name: ee440
-users: []
--- a/cluster-scope/base/user.openshift.io/groups/ee440/kustomization.yaml
+++ b/cluster-scope/base/user.openshift.io/groups/ee440/kustomization.yaml
@@ -1,4 +1,0 @@
-apiVersion: kustomize.config.k8s.io/v1beta1
-kind: Kustomization
-resources:
-    - group.yaml
--- a/cluster-scope/components/resourcequotas/ope-course-quota/resourcequota.yaml
+++ b/cluster-scope/components/resourcequotas/ope-course-quota/resourcequota.yaml
@@ -8,5 +8,5 @@ spec:
     limits.cpu: '1000'
     limits.ephemeral-storage: 30Gi
     limits.memory: 3000000Mi
-    persistentvolumeclaims: '400'
-    requests.storage: 400Gi
+    persistentvolumeclaims: '600'
+    requests.storage: 600Gi
--- a/cluster-scope/overlays/nerc-ocp-obs/feature/odf/externalsecrets/rook-ceph-external-cluster-details.yaml
+++ b/cluster-scope/overlays/nerc-ocp-obs/feature/odf/externalsecrets/rook-ceph-external-cluster-details.yaml
@@ -1,0 +1,16 @@
+apiVersion: external-secrets.io/v1beta1
+kind: ExternalSecret
+metadata:
+  name: rook-ceph-external-cluster-details
+  namespace: openshift-storage
+spec:
+  secretStoreRef:
+    name: nerc-secret-store
+    kind: SecretStore
+  target:
+    name: rook-ceph-external-cluster-details
+  data:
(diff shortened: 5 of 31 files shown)


logging/overlays/nerc-ocp-obs/kustomization.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+resources:
+- ../../base
```
computate: The loki/base/objectbucketclaims/openshift-logging-objectbucketclaim.yaml resource references a `metrics-bucket-storage` StorageClass which is not defined yet in the obs cluster. We should create another PR that sets up a StorageClass, BucketClass, and BackingStore with the right size mentioned in [Setup storage for new Logs Metrics Cluster](https://github.com/nerc-project/operations/issues/240), which for default log storage says it should be `10 Ti`. 

Keep in mind that we shouldn't call it metrics, which was a mistake in the infra cluster. Think about what we should call the StorageClass, BucketClass, and BackingStore for both logs with retention and archived logs for 1 year as described in the issue above.

computate: Remember that when you configure a BackingStore like our `metrics-backing-store`, in the beginning we set it to `storage: 400Gi` and `numVolumes: 3`. Because we kept running out of log storage we had to keep increasing up to `numVolumes: 24`. At 24, we got a different error and through Red Hat Support Cases we figured out that 20 is the max `numVolumes` for a BackingStore, so we bumped it down to 20, [see the commits here to BackingStore](https://github.com/OCP-on-NERC/nerc-ocp-config/commits/main/cluster-scope/overlays/nerc-ocp-infra/metrics-bucket-storage/backingstore.yaml). Also, once you set `storage: 400Gi` in a BackingStore, you cannot change it, it will have no effect in the PVCs if you change it. So when choosing the number of volumes, keep in mind that the `numVolumes * storage == 10Ti` is what we are looking for, and it creates `numVolumes` PVCs in the openshift-storage namespace, [see here](https://console-openshift-console.apps.nerc-ocp-infra.rc.fas.harvard.edu/k8s/ns/openshift-storage/core~v1~PersistentVolumeClaim?name=metrics-backing-store). I think it's good to have more than one volume if we have multiple nodes in the cluster.

computate: Also the LokiStack configures `storageClassName: ocs-external-storagecluster-ceph-rbd`. We want to make sure this is in place in the obs cluster.

logging/base/clusterlogforwarders/instance.yaml
```diff
@@ -4,11 +4,15 @@ metadata:
   name: instance
   namespace: openshift-logging
 spec:
```
computate: Let's revert the changes to instance.yaml because we use patches to override these values for the prod and test cluster to send logs remotely.

computate (changes requested):

schwesig:
@schwesig

logging/overlays/nerc-ocp-obs/externalsecrets/openshift-logging-lokistack-gateway-bearer-token_patch.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: external-secrets.io/v1beta1
+kind: ExternalSecret
+metadata:
+  name: openshift-logging-lokistack-gateway-bearer-token
+  namespace: openshift-logging
+spec:
+  dataFrom:
+    - extract:
+        # Command to extract the JSON pull secret: oc extract secret/pull-secret -n openshift-config --to=-
+        key: nerc/nerc-ocp-obs/lokistack-gateway-bearer-token
```
computate: Let's remove this ExternalSecret in `logging/overlays/nerc-ocp-obs` because the LokiStack will create it.

logging/overlays/nerc-ocp-obs/kustomization.yaml
```diff
@@ -0,0 +1,8 @@
...
+
+patches:
+  - path: externalsecrets/openshift-logging-lokistack-gateway-bearer-token_patch.yaml
```
computate: Let's remove this patch to logging/overlays/nerc-ocp-test because the LokiStack will create this secret.

computate (changes requested):

PR #349, opened 2024-01-22
Add missing admin-acks

larsks:
The config for nerc-ocp-test appears to have lost some admin-acks values;
these are set in the cluster but not in the repository. This commit makes
the repository match.


schwesig (approved):
thanks @larsks for checking into this

computate (approved):

PR #350, opened 2024-01-22
Fix erroneous labels on curator deployment

larsks:
The curator deployment was using labels that are maintained by argocd. This
removes the app.kubernetes.io/instance label and fixes the
app.kubernetes.io/name label.


schwesig (approved):
thanks @larsks for finding this

PR #340, opened 2024-01-15
Create and add ope groups cs210 & ee440 to prod cluster

dystewart:

PR #352, opened 2024-01-25
Rollback changes made in PR #340

dystewart:
https://github.com/OCP-on-NERC/nerc-ocp-config/pull/340/

We will be making manual syncs of students added to their course namespace to these OCP groups. We aren't tracking users in these groups via github so any changes to the group in the live cluster are overwritten by argoCD

We will still be tracking the group manifets themselves albeit in a different ope dedicated repo: https://github.com/OCP-on-NERC/BU-RHOAI


larsks (approved):

schwesig (approved):
ack, looking good

PR #357, opened 2024-01-30
Add http01 letsencrypt ClusterIssuer to prod and obs cluster

### LINKED ISSUE #413: Create http-01 clusterissuer on production cluster
We should create an HTTP-01 cluster-wide certificate issuer on the production cluster because this would allow people to host custom domain names on the cluster.

computate:
A LetsEncrypt ClusterIssuer with an http01 solver is useful for
acquiring non-wildcard certificates for arbitrary domain names without
requiring any DNS credentials. We still plan on relying on the wildcard
default ingress certificate for most common use cases.

Closes nerc-project/operations#413

### DIFF
--- a/cluster-scope/base/cert-manager.io/clusterissuers/letsencrypt-production-http01/clusterissuer.yaml
+++ b/cluster-scope/base/cert-manager.io/clusterissuers/letsencrypt-production-http01/clusterissuer.yaml
@@ -1,0 +1,13 @@
+apiVersion: cert-manager.io/v1
+kind: ClusterIssuer
+metadata:
+  name: letsencrypt-production-http01
+spec:
+  acme:
+    privateKeySecretRef:
+      name: letsencrypt-production-http01
+    server: https://acme-v02.api.letsencrypt.org/directory
+    solvers:
+      - http01:
+          ingress:
+            class: openshift-default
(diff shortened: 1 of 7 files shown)


larsks (changes requested):
Two requests:

1. Please create separate staging and production ClusterIssuers as we do with the DNS-01 issuers.

2. Please update the PR description and commit message to reflect that the primary advantage here is that this allows us to acquire non-wildcard certificates for arbitrary domain names without requiring any DNS credentials. We still plan on relying on the wildcard default ingress certificate for most common use cases.

computate:
>     1. Please create separate staging and production ClusterIssuers as we do with the DNS-01 issuers.
done
>     2. Please update the PR description and commit message to reflect that the primary advantage here is that this allows us to acquire non-wildcard certificates for arbitrary domain names without requiring any DNS credentials. We still plan on relying on the wildcard default ingress certificate for most common use cases.
done

Thanks for the feedback @larsks

larsks (approved):
Approved, although I'm not sure we really need this in our commit messages:

```
    rh-pre-commit.version: 2.0.3
    rh-pre-commit.check-secrets: ENABLED
```

Generally, we try not to clutter things up with details of our local development environment (automated annotations in commit  messages, editor modelines, etc).

computate:
Sure @larsks I disabled Red Hat Pre Commit messages for my NERC repos and pushed up a cleaner commit.

dystewart (approved):

PR #358, opened 2024-01-30
obs cluster default-api-certificate to use http01 issuer

computate:
We set the default API certificate in the obs cluster to use an http01 letsencrypt issuer without requiring any DNS credentials.


schwesig (approved):
good to go

schwesig:
addition to https://github.com/OCP-on-NERC/nerc-ocp-config/pull/357

dystewart (approved):

PR #359, opened 2024-01-31
Correct issuer reference for obs default-api-certificate

larsks:
The default-api-certificate manifest had:

    spec:
      issuerRef:
        name: letsencrypt-production-http01
        kind: Issuer

But we are using a ClusterIssuer, so we need:

    spec:
      issuerRef:
        name: letsencrypt-production-http01
        kind: ClusterIssuer


computate (approved):

larsks:
This PR is no longer necessary.

PR #360, opened 2024-01-31
Update X Small Resource size

IsaiahStapleton:
We need to update the X Small Resource size to use a smaller amount of resources in order to keep class costs low. These are the resource constraints that Orran asked for: https://docs.google.com/document/d/1vQ6BtcBi2kPh1_D4or1OTxcgrRun9qnKR-Ew2Yz6bxQ/edit


schwesig (approved):
looking good, as discussed in the call with Orran

dystewart (approved):

PR #361, opened 2024-01-31
Use ClusterIssuers instead of namedspaced Issusers on obs cluster

larsks:
This pull requests gather together several commits that impact our cert-manager configuration. Primarily, we are (a) introducing two ClusterIssuer resources into the repository (to be used on clusters on which we are not running arbitrary user workloads), and (b) we are making use of these ClusterIssuers on the observability cluster.

- Revert "default-api-certificate to use http01 issuer (#358)"
- Fix typo in staging issuer configuration
- Add dns-01 based ClusterIssuer
- Replace namespaced Issuers with ClusterIssuers on obs cluster


computate (approved):

schwesig (approved):
@larsks thanks for all your help here.

PR #362, opened 2024-02-01
Update obs namedCertificate to api.obs.nerc.mghpcc.org

computate:
The namedCertificate was pointing to a domain that doesn't exist
api.nerc-ocp-obs.rc.fas.harvard.edu


bnshr (approved):

schwesig (approved):
worked on it in the call. sounds good to me.

larsks (approved):

PR #364, opened 2024-02-01
Patch server name in ClusterSecretStore

larsks:
The ClusterSecretStore was originally created for the infra cluster, where
we access the vault via its in-cluster address. We need to modify that when
using the manifest on other clusters.

Probably we should swap our logic at some point in the future (that is,
default to using the externally visible name), but right now it's the same
amount of work either way.


schwesig (approved):
👍

PR #365, opened 2024-02-02
Fake metrics endpoint to make ODF happy in obs cluster

computate:
While the "monitoring-endpoint" configuration in OCS 4.8 was optional,
it appears to be required in 4.10, and ODF refuses to deploy if it
can't reach the endpoint.

This commit provides a fake metrics endpoint to make the ODF operator
happy.

### DIFF
--- a/fake-metrics-server/overlays/nerc-ocp-obs/kustomization.yaml
+++ b/fake-metrics-server/overlays/nerc-ocp-obs/kustomization.yaml
@@ -1,0 +1,5 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+namespace: openshift-storage
+resources:
+- ../nerc-ocp-infra


schwesig (approved):
🙏

dystewart (approved):

larsks:
This looks fine, but before you merge it, please see what happens if you simply remove the `monitoring-endpoint` section from the `rook-ceph-external-cluster-details` secret. This hack was necessary for an earlier version of ODF; it would be good to verify that it is still necessary.

computate:
Looks like it's still required @larsks , based on this updated error: `Error while reconciling: Unable to retrieve "monitoring-endpoint"`

```yaml
    - lastHeartbeatTime: '2024-02-02T20:29:42Z'
      lastTransitionTime: '2024-01-25T13:59:24Z'
      message: >-
        Error while reconciling: Unable to retrieve "monitoring-endpoint"
        external resource
      reason: ReconcileFailed
      status: 'False'
      type: ReconcileComplete
```

PR #368, opened 2024-02-08
Configure bond0 on obs cluster

larsks:
Create the bond0 interface with members eno1 and eno2. This will allow us
to attach the storage vlan to bond0 in a future commit.

NB: For this to work, we had to apply the following custom selinux policy
to the worker nodes:

    module local 1.0;

    require {
            type openvswitch_t;
            type tmpfs_t;
            class file { create open read write };
    }

    allow openvswitch_t tmpfs_t:file { create open read write };

This appears unrelated to this commit, and may represent an OpenShift bug,
but we're a few revisions behind the latest 4.14.x release so it's possible
it has already been fixed.


computate (approved):

bnshr (approved):

computate:
@larsks this looks good to me. @schwesig is out today, so feel free to merge it.

schwesig (approved):
Following the long thread in slack. Thanks for figuring this out. 🙏

PR #372, opened 2024-02-09
Upgrade loki-operator to 5.7 on obs, leave logging operator for now

computate:
Changing this PR to just focus on the obs loki-operator as the beginning of some follow up PRs to update step by step.

old idea:
```
The Loki Operator Subscription on the obs cluster is unable to install
because of an error: constraints not satisfiable: no operators found in
channel stable-5.6 of package loki-operator in the catalog referenced by
subscription loki-operator, subscription loki-operator exists.

We will upgrade Loki on the infra cluster and obs cluster because
version 5.6 is no longer supported.
```

### DIFF
--- a/cluster-scope/overlays/nerc-ocp-obs/kustomization.yaml
+++ b/cluster-scope/overlays/nerc-ocp-obs/kustomization.yaml
@@ -110,3 +110,10 @@ patches:
             kubernetes:
               mountPath: kubernetes/nerc-ocp-obs
           server: https://vault-ui-vault.apps.nerc-ocp-infra.rc.fas.harvard.edu/
+- target:
+    kind: Subscription
+    name: loki-operator
+  patch: |
+    - op: replace
+      path: /spec/channel
+      value: stable-5.7


cluster-scope/base/operators.coreos.com/subscriptions/loki-operator/subscription.yaml
```diff
@@ -4,7 +4,7 @@ metadata:
   name: loki-operator
   namespace: openshift-operators-redhat
 spec:
-  channel: stable-5.6
+  channel: stable-5.8
```
larsks: This is changing the base version of the loki operator on *all* clusters. Do we have it deployed anywhere other than the obs cluster? In particular, is it running on the prod cluster, and is 5.8 compatible with the OCP version there?

computate: @larsks @schwesig @bnshr  It's deployed on the `infra` cluster, and we are trying to deploy it on the `obs` cluster. I also realized that the Loki Operator is also closely related to the Logging Operator, which is configured to be `channel: stable`. Should we make Loki and Logging Operator both `channel: stable` so that we always have stable/supported logging, or increment versions?

schwesig: yes to `stable`,  from my side.

schwesig: [larsks](https://massopencloud.slack.com/archives/C027TDE52TZ/p1707490673081909?thread_ts=1707489261.502159&cid=C027TDE52TZ)
I don't know; if stable will change versions (e.g., move from 5.7 to 5.8 automatically), that seems like it might introduce breaking changes. I would probably stick with stable-5.8, and only explicitly upgrade to a new minor release.

schwesig: call with @computate we think going with 5.7 on obs and then upgrade there with some smaller PRs would be the better solution

computate: @larsks we have updated the commit to only patch the loki-operator version on the obs cluster to `5.7`. Please review again.

schwesig (approved):
check, worked on it in our Monday call

larsks (approved):

PR #375, opened 2024-02-13
Remove rolebinding missing role

tssala23:
This Cluster Role Binding is binding a role that does not exist on any of
the clusters.


computate:
@tssala23  I think instead of deleting the `open-cluster-management:cluster-manager-admin` ClusterRoleBinding, we should make it specific to the infra cluster, where this Role is defined.

computate (approved):

larsks (approved):

PR #378, opened 2024-02-15
Refactor nerc-ops rbac

larsks:
Previously, users in the nerc-ops group received permissions through a
matrix of clusterroles and clusterrolebindings. This was clunky, and
adding new permissions meant creating new roles and rolebindings.

With this commit, we replace all of the *nerc-ops* clusterroles and
clusterrolebindings with:

- A single clusterrolebinding, `nerc-ops`.
- A clusterrole `nerc-ops` that makes us of aggregation rules [1] to
  act just like the "view" and "cluster-reader" roles, while also
  making it easy to add nerc-ops specific permissions.
- A suite of `nerc-ops-*` cluster roles that add additional rules to
  the `nerc-ops` role, specifically permitting members of `nerc-ops`
  to:

  - List secrets
  - `kubectl exec` into pods
  - Impersonate `system:admin` using the `--as` option

[1]: https://kubernetes.io/docs/reference/access-authn-authz/rbac/#aggregated-clusterroles


dystewart (approved):

schwesig (approved):
great idea! continously improving.

PR #380, opened 2024-02-19
Upgrade and refactor vault deployment

larsks:
This pull request replaces our previous vault deployment with one managed using the bank-vaults [vault operator](https://github.com/bank-vaults/vault-operator). The vault operator handles automatically unsealing the vault when necessary (nerc-project/operations#446). At the time we first installed vault, the operator had been abandoned (https://github.com/banzaicloud/bank-vaults), but development has since resumed under [a new organization](https://github.com/bank-vaults).

This deployment uses the current vault release, 1.15.5 (nerc-project/operations#439).

I have organized the vault deployment into three separate "applications":

- `vault/operator` is just the vault operator. It does not create any vault instances.
- `vault/instance` is the actual vault instance.
- `vault/backup-job` is the backup job.

**NB**: 10456 lines of this change are the vault CRD. I should note that we're installing the operator manifests directly because there is no maintained version of this operator in OperatorHub.


vault/backup-job/base/trigger/clusterrolebinding.yaml
```diff
@@ -9,4 +9,4 @@ roleRef:
 subjects:
 - kind: ServiceAccount
   name: backup-job
-  namespace: vault
+  namespace: default
```
larsks, line 12: This is intentional; when using the namespace `default`, kustomize will update the value when there is a `namespace:` set in `kustomization.yaml`.

schwesig (approved):
wow, a big clean up. Thanks Lars.

vault/operator/base/src/.gitignore
dystewart, line 1: Is this file included intentionally?

larsks: Yes. It prevents accidentally adding the `charts` directory that gets created when you run the Makefile in the parent directory.

dystewart:
LGTM

dystewart (approved):

dystewart:
Good to see the bank vaults operator is being maintained again

PR #381, opened 2024-02-19
nerc obs admins sudoer

computate:
**Grant sudoer to nerc-obs-admins on obs cluster**

The NERC Observability admins should have sudoer access on the obs cluster to access secrets in openshift-logging and open-cluster-management-observability namespace.

### DIFF
--- a/cluster-scope/base/rbac.authorization.k8s.io/clusterrolebindings/nerc-obs-admins-sudoer/clusterrolebinding.yaml
+++ b/cluster-scope/base/rbac.authorization.k8s.io/clusterrolebindings/nerc-obs-admins-sudoer/clusterrolebinding.yaml
@@ -1,0 +1,12 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: nerc-obs-admins-sudoer
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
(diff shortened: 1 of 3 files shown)


schwesig (approved):
looking good to me, as discussed and created in the call.
having @larsks to take a look, after he had set up the way more clear structure in authorization logic

larsks (changes requested):
The PR description says...

> They also need access to secrets in openshift-logging and open-cluster-management-observability namespace.

...but I don't see any RBAC here granting secrets access. Are we missing rbac in this pr, or should that comment be moved to a future PR?

larsks:
> looking good to me, as discussed and created in the call. having @larsks to take a look, after he had set up the way more clear structure in authorization logic

I would like to handle this more like we we're handling nerc-ops permissions in https://github.com/OCP-on-NERC/nerc-ocp-config/pull/378, but we can go ahead and approve it for now and make changes later.

The goal here is pretty much to create a cluster-local "ops" group. It would be great if we could make that a generic, re-usable configuration, so that we wouldn't have to write the same set of manifests for every cluster.

computate:
@larsks I updated the commit and description because access to the logs is unrelated to the sudoers role, I wasn't thinking clearly there. 
The sudoers role should allow them to grab the secrets to query logs and metrics via API to discover better ways to query them.

larsks (approved):

PR #383, opened 2024-02-19
[DO NOT MERGE] Create a generic, re-usable ops role

larsks:
@schwesig @computate this is the idea I was talking aboutin #381. We create a generic "ops" role, and then use kustomize patching magic to re-create our existing "nerc-ops" role as well as a new "local-ops" role, and then we make use of that "local-ops" role on the obs cluster.

This was a quick hack so I'm not actually suggesting we merge this now, but I wanted to put it out here for folks to see.

### DIFF
--- a/cluster-scope/base/rbac.authorization.k8s.io/clusterrolebindings/ops/clusterrolebinding.yaml
+++ b/cluster-scope/base/rbac.authorization.k8s.io/clusterrolebindings/ops/clusterrolebinding.yaml
@@ -1,0 +1,14 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  labels:
+    nerc.mghpcc.org/bundle: ops-rbac
+  name: ops
+roleRef:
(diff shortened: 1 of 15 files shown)


schwesig:
@larsks this looks great, and I would love to follow that. But without your help I wouldn't be able to create this scheme by myself.
Thanks for the quick draw and concept.

larsks:
@schwesig 

> But without your help I wouldn't be able to create this scheme by myself.

That's okay, in retrospect this is dumb and you should not aspire to create it :smile:. We can of course accomplish the same thing with a single ClusterRoleBinding manifest:

```
kind: ClusterRoleBinding
metadata:
  name: nerc-ops-on-obs
roleRef:
  kind: ClusterRole
  name: nerc-ops
subjects:
- apiGroup: rbac.authorization.k8s.io
  kind: Group
  name: nerc-obs-admins
```

That is, we can just bind another group to the `nerc-ops` role on a given cluster.

That's much simpler and accomplishes the same thing :).

larsks:
In fact, that's such a simple solution I'm just going to make a PR for it.

PR #384, opened 2024-02-19
Grant nerc-ops role to members of nerc-obs-admins on obs cluster

larsks:
This grants members of the`nerc-obs-admin` group the same privileges as
members of `nerc-ops`, but only on the `nerc-ocp-obs` cluster. That means
they can:

- View most cluster resources, including secrets
- Impersonate `system:admin` ("sudo")
- Use `kubectl exec`


computate (approved):

schwesig (approved):
thanks for the update

PR #389, opened 2024-02-21
Fix attribute name in nfd manifest

larsks:
According to `kubectl explain nodefeaturediscovery.spec`, `topologyUpdater`
should be spelled `topologyupdater`. The mismatch is making this resource
show up as out-of-sync in argocd.

### DIFF
--- a/nfd-operator/base/nodefeaturediscoveries/nfd-instance.yaml
+++ b/nfd-operator/base/nodefeaturediscoveries/nfd-instance.yaml
@@ -9,7 +9,7 @@ spec:
     image: registry.redhat.io/openshift4/ose-node-feature-discovery:v4.11
     imagePullPolicy: Always
     servicePort: 0
-  topologyUpdater: false
+  topologyupdater: false
   workerConfig:
     configData: |
       core:


nfd-operator/base/nodefeaturediscoveries/nfd-instance.yaml
```diff
@@ -9,7 +9,7 @@ spec:
     image: registry.redhat.io/openshift4/ose-node-feature-discovery:v4.11
     imagePullPolicy: Always
     servicePort: 0
-  topologyUpdater: false
+  topologyupdater: false
```
naved001, line 12: Your commit message says "`topologyUpdated` should be spelled `topologyupdated`" but here it's ending in "er" (Updater -> updater). Which is it?

larsks: My commit message is wrong. Fixing it...

naved001 (approved):

computate (approved):

schwesig (approved):

PR #397, opened 2024-03-08
ocp-test: add mlx5_core udev rules for A100 hosts

jtriley:
This adds a machine config to install the udev rules needed for the NICs on the Lenovo SD530 A100 hosts. We will be adding one of these hosts to the test cluster to experiment with MIG support (see https://github.com/nerc-project/operations/issues/468)


dystewart (approved):

schwesig (approved):
thanks @jtriley

PR #398, opened 2024-03-08
upgrade Loki on obs and switch Logging Operator from stable to stable-5.8 for all clusters

computate:
- **Upgrade Loki operator to stable-5.8 on obs cluster**
- **Change Logging operator version from stable to stable-5.8**

### DIFF
--- a/cluster-scope/base/operators.coreos.com/subscriptions/cluster-logging/subscription.yaml
+++ b/cluster-scope/base/operators.coreos.com/subscriptions/cluster-logging/subscription.yaml
@@ -4,7 +4,7 @@ metadata:
   name: cluster-logging
   namespace: openshift-logging
 spec:
-  channel: stable
+  channel: stable-5.8
   installPlanApproval: Automatic
   name: cluster-logging
   source: redhat-operators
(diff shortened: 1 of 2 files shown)


schwesig (approved):
thanks

cluster-scope/base/operators.coreos.com/subscriptions/cluster-logging/subscription.yaml
```diff
@@ -4,7 +4,7 @@ metadata:
   name: cluster-logging
   namespace: openshift-logging
 spec:
-  channel: stable
+  channel: stable-5.8
```
larsks, line 7: It looks like we're making the change to `stable-5.8` globally, rather than just on the obs cluster (which is the verbage in the pr description). Is this okay?

computate: @larsks the title of the PR was confusing, so I renamed it to "upgrade Loki on obs and switch Logging Operator from stable to stable-5.8 for all clusters"

larsks (approved):

PR #399, opened 2024-03-11
Allow nerc-ops view access on monitoring resources

larsks:
Allow members of the nerc-ops group to view/list resources in the
monitoring.coreos.com api group. This is necessary to be able to run e.g.
the `noobaa status` command.


computate (approved):

schwesig (approved):
👍

PR #413, opened 2024-03-21
Add vault configuration for NERC internal CA

### LINKED ISSUE #424: We need a way to generate locally valid certificates to enable vault access from internal clusters
In order for a cluster to successfully access the vault instance on nerc-ocp-infra, the vault service must trust the cluster API certificate. For public-facing clusters, we are able to generate certificates that are globally trusted using Lets Encrypt.

For the infra cluster, we manually issue an InCommon certificate.

For other internal clusters, such as nerc-ocp-test, we are not able to use Lets Encrypt because:

1. The clusters have no public address so they cannot use HTTP-01 challenges, and
2. There is no public facing DNS for the `rc.fas.harvard.edu` domain

This means that it is not currently possible for these clusters to access the vault. There are a number of possible solutions:

- If we implement #214 we can use Lets Encrypt to generate globally valid certificates
- We can manually issue InCommon certificates for all internal clusters
- We can create our own CA on nerc-ocp-infra that is trusted by the vault, and use that to issue certificates for internal clusters.

larsks:
Enable Vault PKI support [1] at pki-nerc-root. Certificates issued by this
CA will be trusted by the vault, so that test/dev clusters (or other
clusters that don't require a globally trusted certificate) can have vault
access.

With these changes we can manually issue certificates. A future change will
enable ACME API support so that we can automatically generate internal
certificates.

Closes: nerc-project/operations#424

[1]: https://developer.hashicorp.com/vault/docs/secrets/pki


schwesig (approved):
Thanks for taking care of this topic 🙏

computate (approved):

PR #419, opened 2024-04-02
Add Serverless/Knative CRDs aggregate-to-edit ClusterRole

### LINKED ISSUE #519: issues while using KNative
From @Milstein We have some outstanding issues while using KNative:

1.  When I go into the "Deployment" of a serverless function, I add environment variables through the "Single Values" interface. After saving however, the page refreshes and the environment variable is not present. Similar result if added through the YAML. Neither approach stores the environment variable.

2. I was trying to create an event source for KNative and I run into the following error message.
Default alert:Event source cannot be created

You do not have access for Event Source in this project.

This is by going through Administrator > Eventing > Create > Event Source > PingSource

computate:
- **Add Serverless/Knative CRDs aggregate-to-edit ClusterRole**
Grant Users access to create and edit Knative CRDs in their namespace like PingSource, ApiServerSource, ContainerSource, SinkBinding, Channel, Broker. 

Closes https://github.com/nerc-project/operations/issues/519

### DIFF
--- a/cluster-scope/base/rbac.authorization.k8s.io/clusterroles/knative-edit/clusterrole.yaml
+++ b/cluster-scope/base/rbac.authorization.k8s.io/clusterroles/knative-edit/clusterrole.yaml
@@ -1,0 +1,22 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: knative-edit
+  labels:
+    rbac.authorization.k8s.io/aggregate-to-edit: "true"
+rules:
+  - apiGroups:
+      - 'sources.knative.dev/v1'
+      - 'serving.knative.dev/v1'
+      - 'messaging.knative.dev/v1'
+      - 'eventing.knative.dev/v1'
+    resources:
+      - '*'
+    verbs:
+      - create
+      - delete
+      - update
+      - patch
+      - get
+      - watch
+      - list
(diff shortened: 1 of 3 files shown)


schwesig (approved):
LGTM thanks

larsks (changes requested):
These changes have nothing to do with RBAC and should not be part of the allow-edit-rbac role.

If these are "unprivileged" operations (that is, we want them to be generally available), create a new clusterrole and label it with:

```
labels:
  rbac.authorization.k8s.io/aggregate-to-edit: "true"
```

larsks (changes requested):
These changes have nothing to do with RBAC and should not be part of the allow-edit-rbac role.

If these are "unprivileged" operations (that is, we want them to be generally available), create a new clusterrole and label it with:

```
labels:
  rbac.authorization.k8s.io/aggregate-to-edit: "true"
```

This will make them available to anyone with the `edit` role.

computate:
Thanks for the feedback @larsks, I have moved the rules to a new `knative-edit` ClusterRole with `aggregate-to-edit`. Please review again. This PR is to allow users where the Serverless Operator is deployed to create/delete/update/patch/get/watch/list these related knative CRDs in their namespace with the edit role: 

- 'sources.knative.dev/v1'
- 'serving.knative.dev/v1'
- 'messaging.knative.dev/v1'
- 'eventing.knative.dev/v1'

larsks (approved):

PR #425, opened 2024-04-05
fix: repeat_interval cannot be zero

schwesig:
fix: repeat_interval cannot be zero

### DIFF
--- a/cluster-scope/base/external-secrets.io/externalsecrets/open-cluster-management-observability-alertmanager-config/externalsecret.yaml
+++ b/cluster-scope/base/external-secrets.io/externalsecrets/open-cluster-management-observability-alertmanager-config/externalsecret.yaml
@@ -92,7 +92,7 @@ spec:
               - receiver: slack-notifications-prod-rhods-ope
                 group_by: ['...']
                 group_wait: 0s
-                repeat_interval: 0s
+                # repeat_interval cannot be zero, group_interval cannot be zero
                 matchers:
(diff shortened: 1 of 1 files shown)


cluster-scope/base/external-secrets.io/externalsecrets/open-cluster-management-observability-alertmanager-config/externalsecret.yaml
```diff
@@ -92,7 +92,7 @@ spec:
               - receiver: slack-notifications-prod-rhods-ope
                 group_by: ['...']
                 group_wait: 0s
-                repeat_interval: 0s
+                # repeat_interval cannot be zero, group_interval cannot be zero
```
naved001, line 95: what does it default to if it's not set here at all?

schwesig: it is 4 hours. https://github.com/prometheus/alertmanager/issues/860#issuecomment-467585781
and I need to get it down to less for my specific case.
but right now, alert manager pod is not starting at all, and I want at least the other alerst running as usual.
and take care of my special wishes for the alert after the weekend.

naved001 (approved):

DanNiESh (approved):

schwesig:
fyi:
[How do you guys validate the alertmanager.yaml file, is there a tool for this purpose. #2926](https://github.com/prometheus/alertmanager/issues/2926)

PR #439, opened 2024-05-01
obs ingress cert

computate:
- **Fixing the ClusterIssuer for ingress cert on obs**


larsks (approved):

computate:
closes nerc-project/operations#554

PR #445, opened 2024-05-09
obs keycloak permissions

### LINKED ISSUE #493: Implement a filtering proxy for controlling access to Prometheus metrics
A core goal of our observability work is to be able to expose logs and metrics for research purposes. In order to accomplish this, we need a mechanism for controlling access to logs and metrics such that we can grant access to *some* data without granting access to *all* data.

@computate has already started looking at this in the context of #453. We had a conversation earlier today and talked about a couple of different ways of implementing access control:

1. Query filtering

    In this model, we filter the query provided by the user, for example by enforcing a specific label on queries. Existing projects that follow this model include [prom-label-proxy](https://github.com/prometheus-community/prom-label-proxy) and [prometheus-filter-proxy](https://github.com/hoffie/prometheus-filter-proxy).

2. Response filtering

    In this model, we permit all queries, but then we filter the response to include only metrics that include specific labels. In some ways this is simpler than query filtering, because it doesn't require parsing or modifying the query, and the format of the return data is relatively straightforward.

I suggested to @computate that option 2 might be the better option, but given that all the prior art seems to favor option 1 I am rethinking my position :).

Neither of the above solutions implements any sort of authorization. In order to provide that, we could place an authorization in proxy in front of the service -- something like [kube-rbac-proxy](https://github.com/brancz/kube-rbac-proxy) or perhaps [prom-authzed-proxy](https://github.com/authzed/prom-authzed-proxy).

Our ultimate goal is to achieve something like the [LBAC](https://grafana.com/docs/enterprise-metrics/latest/tenant-management/lbac/) (label based access control) feature available in Grafana Enterprise.

computate:
- **Upgrade Keycloak operator version from v22 to latest v24**
- **Adding prom-keycloak-proxy to obs cluster**
- Adding a new `nerc` client to Keycloak to contain all of the authorization resources, scopes, policies, and permissions to query against with prom-keycloak-proxy. 
- Add `ai4cloudops` and `oct` clients linked to their client permissions in the `nerc` client authorization configuration.
- Add the prom-keycloak-proxy Deployment, Service, Route, and ExternalSecrets already configured in vault to connect to Keycloak and ACM Observatorium on the infra cluster. 
- Find the [repo and docs about the prom-keycloak-proxy here](https://github.com/nerc-images/prom-keycloak-proxy).
- See the [keycloak-permissions-operator in ](https://operatorhub.io/operator/keycloak-permissions-operator)[operatorhub.io](http://operatorhub.io/), and it's also available in all OpenShift Cluster Operator Hubs.
- Find the [repo and docs about the keycloak-permissions-operator here](https://github.com/nerc-images/keycloak-permissions-operator).

closes nerc-project/operations#493

### DIFF
--- a/keycloak/overlays/nerc-ocp-obs/keycloakauthorizations/nerc/keycloakauthorization.yaml
+++ b/keycloak/overlays/nerc-ocp-obs/keycloakauthorizations/nerc/keycloakauthorization.yaml
@@ -16,13 +16,37 @@ spec:
   realm:
     id: NERC
     clients:
-      - id: ai4cloudops
+      - id: nerc
         authorizationSettings:
+          scopes:
+
+            - id: namespace-all
+              name: all namespaces
+              displayName: all namespaces
+            - id: namespace-rhods-notebooks
+              name: rhods-notebooks
+              displayName: rhods-notebooks namespace
+
+            - id: cluster-all
+              name: all clusters
+              displayName: all clusters
+            - id: cluster-nerc-ocp-infra
+              name: nerc-ocp-infra
+              displayName: nerc-ocp-infra cluster
+            - id: cluster-nerc-ocp-obs
+              name: nerc-ocp-obs
+              displayName: nerc-ocp-obs cluster
+            - id: cluster-nerc-ocp-prod
+              name: nerc-ocp-prod
+              displayName: nerc-ocp-prod cluster
+            - id: cluster-nerc-ocp-test
+              name: nerc-ocp-test
+              displayName: nerc-ocp-test cluster
+
           resources:
             - name: cluster
               displayName: cluster
               scopes:
-
                 - id: cluster-all
                   name: all clusters
                   displayName: all clusters
@@ -50,6 +74,16 @@ spec:
                   displayName: rhods-notebooks namespace
 
           policies:
+
+            - id: group-nerc-ops
+              name: group-nerc-ops
+              logic: POSITIVE
+              decisionStrategy: UNANIMOUS
+              type:
+                group:
+                  groups:
+                    - id: nerc-ops
+
             - id: client-ai4cloudops
               name: client-ai4cloudops
               logic: POSITIVE
@@ -66,39 +100,43 @@ spec:
                 group:
                   groups:
                     - id: nerc-ai4cloudops
-            - id: group-nerc-ops
-              name: group-nerc-ops
+
+            - id: client-oct
+              name: client-oct
               logic: POSITIVE
               decisionStrategy: UNANIMOUS
               type:
-                group:
-                  groups:
-                    - id: nerc-ops
+                client:
+                  clients:
+                    - oct
 
           permissions:
+
             - name: group-nerc-ops-cluster-all
-              client: ai4cloudops
               policy: group-nerc-ops
               resource: cluster
             - name: group-nerc-ops-namespace-all
-              client: ai4cloudops
               policy: group-nerc-ops
               resource: namespace
+
             - name: group-nerc-ai4cloudops-namespace-all
-              client: ai4cloudops
               policy: group-nerc-ai4cloudops
               resource: namespace
             - name: client-ai4cloudops-namespace-all
-              client: ai4cloudops
               policy: client-ai4cloudops
               resource: namespace
             - name: group-nerc-ai4cloudops-cluster-nerc-ocp-prod
-              client: ai4cloudops
               policy: group-nerc-ai4cloudops
               resource: cluster
               scope: nerc-ocp-prod
             - name: client-ai4cloudops-cluster-nerc-ocp-prod
-              client: ai4cloudops
               policy: client-ai4cloudops
               resource: cluster
               scope: nerc-ocp-prod
+
+            - name: client-oct-namespace-all
+              policy: client-oct
+              resource: namespace
+            - name: client-oct-cluster-all
+              policy: client-oct
+              resource: cluster
--- a/prom-keycloak-proxy/overlays/nerc-ocp-obs/deployments/prom-keycloak-proxy/deployment.yaml
+++ b/prom-keycloak-proxy/overlays/nerc-ocp-obs/deployments/prom-keycloak-proxy/deployment.yaml
@@ -1,0 +1,73 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: prom-keycloak-proxy
+  namespace: prom-keycloak-proxy
+spec:
+  replicas: 1
+  revisionHistoryLimit: 0
+  strategy:
+    type: RollingUpdate
+  template:
+    metadata:
+      name: prom-keycloak-proxy
+    spec:
+      containers:
+        - name: prom-keycloak-proxy
+          envFrom:
+            - secretRef:
+                name: prom-keycloak-proxy-secrets
+          env:
+            - name: PROXY_AUTH_REALM
+              value: NERC
+            - name: PROXY_AUTH_BASE_URL
+              value: https://keycloak.apps.obs.nerc.mghpcc.org
+            - name: PROXY_AUTH_TLS_VERIFY
+              value: 'true'
+            - name: PROXY_PROMETHEUS_CA_CRT
+              value: /opt/prom-keycloak-proxy-certs/ca.crt
+            - name: PROXY_PROMETHEUS_TLS_CRT
+              value: /opt/prom-keycloak-proxy-certs/tls.crt
+            - name: PROXY_PROMETHEUS_TLS_KEY
+              value: /opt/prom-keycloak-proxy-certs/tls.key
+          image: quay.io/nerc-images/prom-keycloak-proxy:1.0.1
+          imagePullPolicy: Always
+          livenessProbe:
+            exec:
+              command:
+                - /usr/bin/curl
+                - http://localhost:8080
+            failureThreshold: 6
+            initialDelaySeconds: 10
+            periodSeconds: 10
+            successThreshold: 1
+            timeoutSeconds: 1
+          ports:
+            - containerPort: 8080
+              name: http
+              protocol: TCP
+          readinessProbe:
+            exec:
+              command:
+                - /usr/bin/curl
+                - http://localhost:8080
+            failureThreshold: 6
+            periodSeconds: 10
+            successThreshold: 1
+            timeoutSeconds: 1
+          resources:
+            limits:
+              memory: 1Gi
+              cpu: '1'
+          volumeMounts:
+            - mountPath: /opt/prom-keycloak-proxy-certs
+              name: prom-keycloak-proxy-certs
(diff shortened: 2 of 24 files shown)


schwesig (approved):

keycloak/overlays/nerc-ocp-obs/keycloakauthorizations/nerc/keycloakauthorization.yaml
```diff
@@ -16,13 +16,37 @@ spec:
...
+            - id: namespace-all
+              name: all namespaces
+              displayName: all namespaces
+            - id: namespace-rhods-notebooks
+              name: rhods-notebooks
+              displayName: rhods-notebooks namespace
+
+            - id: cluster-all
+              name: all clusters
+              displayName: all clusters
```
larsks, line 32: What is the difference between a `name` and a `displayName`?

prom-keycloak-proxy/overlays/nerc-ocp-obs/deployments/prom-keycloak-proxy/deployment.yaml
```diff
@@ -0,0 +1,99 @@
...
+    deployment: prom-keycloak-proxy
+  name: prom-keycloak-proxy
+  namespace: prom-keycloak-proxy
+spec:
+  replicas: 1
+  revisionHistoryLimit: 0
+  selector:
+    matchLabels:
+      app: prom-keycloak-proxy
+      deployment: prom-keycloak-proxy
```
larsks: By preference I would remove `metadata.labels` and `spec.selector.matchLabels` from the deployment manifest itself and add them via the `commonLabels` directive in the `kustomization.yaml` file. This ensures that they always match.

prom-keycloak-proxy/overlays/nerc-ocp-obs/deployments/prom-keycloak-proxy/deployment.yaml
```diff
@@ -0,0 +1,99 @@
...
+  strategy:
+    activeDeadlineSeconds: 21600
+    recreateParams:
+      timeoutSeconds: 600
+    resources: {}
```
larsks: Are you explicitly setting these to non-default values (I'm asking because I don't know off the top of my head what the default values are)?

prom-keycloak-proxy/overlays/nerc-ocp-obs/deployments/prom-keycloak-proxy/deployment.yaml
```diff
@@ -0,0 +1,99 @@
...
+    type: Recreate
+  template:
+    metadata:
+      labels:
+        app: prom-keycloak-proxy
+        deployment: prom-keycloak-proxy
```
larsks: Ditto re: using `commonLabels` in `kustomization.yaml`

prom-keycloak-proxy/overlays/nerc-ocp-obs/deployments/prom-keycloak-proxy/deployment.yaml
```diff
@@ -0,0 +1,99 @@
...
+              value: /opt/prom-keycloak-proxy-certs/tls.crt
+            - name: PROXY_PROMETHEUS_TLS_KEY
+              value: /opt/prom-keycloak-proxy-certs/tls.key
+          image: quay.io/nerc-images/prom-keycloak-proxy:1.0.0
+          imagePullPolicy: Always
+          livenessProbe:
+            exec:
+              command:
+                - /usr/bin/curl
+                - http://localhost:8080
```
larsks, line 39: An [http liveness probe](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-a-liveness-http-request) would be simpler, and would work correctly even if `curl` wasn't available in the image.

prom-keycloak-proxy/overlays/nerc-ocp-obs/deployments/prom-keycloak-proxy/deployment.yaml
```diff
@@ -0,0 +1,99 @@
...
+          name: sso
+          ports:
+            - containerPort: 8080
+              name: http
+              protocol: TCP
+          readinessProbe:
+            exec:
+              command:
+                - /usr/bin/curl
+                - http://localhost:8080
```
larsks, line 53: Ditto re: using an http probe

prom-keycloak-proxy/overlays/nerc-ocp-obs/deployments/prom-keycloak-proxy/deployment.yaml
```diff
@@ -0,0 +1,99 @@
...
+            - name: PROXY_AUTH_CLIENT_ID
+              valueFrom:
+                secretKeyRef:
+                  key: PROXY_AUTH_CLIENT_ID
+                  name: prom-keycloak-proxy-secrets
+            - name: PROXY_AUTH_CLIENT_SECRET
+              valueFrom:
+                secretKeyRef:
+                  key: PROXY_AUTH_CLIENT_SECRET
+                  name: prom-keycloak-proxy-secrets
```
larsks: If the key in the Secret matches the environment variable name, you can replace these with an `envFrom` directive:

```suggestion
        - envFrom:
            secretRef:
              name: prom-keycloak-proxy-secrets
          env:
```

prom-keycloak-proxy/overlays/nerc-ocp-obs/services/prom-keycloak-proxy/service.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: prom-keycloak-proxy
+  labels:
+    app: prom-keycloak-proxy
+    deployment: prom-keycloak-proxy
```
larsks: Using `commonLabels` in your `kustomization.yaml` will also keep these labels and `spec.selector` in sync with the labels in the deployment.

larsks (changes requested):

keycloak/overlays/nerc-ocp-obs/keycloakauthorizations/nerc/keycloakauthorization.yaml
computate: If I don't specify a displayName for all scopes, despite it being the same as the name, it looks ugly in the Keycloak UI. 
image

prom-keycloak-proxy/overlays/nerc-ocp-obs/deployments/prom-keycloak-proxy/deployment.yaml
computate: Because all requests to this proxy need to be authenticated, the http liveness probes fail with: 
`Readiness probe failed: HTTP probe failed with statuscode: 401`
Which is why I use curl instead, which doesn't complain about the status code.

prom-keycloak-proxy/overlays/nerc-ocp-obs/deployments/prom-keycloak-proxy/deployment.yaml
computate: I applied this suggestion, thanks.

prom-keycloak-proxy/overlays/nerc-ocp-obs/deployments/prom-keycloak-proxy/deployment.yaml
computate: You are right, I removed them.

prom-keycloak-proxy/overlays/nerc-ocp-obs/deployments/prom-keycloak-proxy/deployment.yaml
computate: I applied this suggestion, thanks.

prom-keycloak-proxy/overlays/nerc-ocp-obs/deployments/prom-keycloak-proxy/deployment.yaml
computate: I applied this suggestion, thanks.

prom-keycloak-proxy/overlays/nerc-ocp-obs/services/prom-keycloak-proxy/service.yaml
computate: I applied this suggestion, thanks.

prom-keycloak-proxy/overlays/nerc-ocp-obs/deployments/prom-keycloak-proxy/deployment.yaml
computate: Because all requests to this proxy need to be authenticated, the http liveness probes fail with:
`Readiness probe failed: HTTP probe failed with statuscode: 401`
Which is why I use curl instead, which doesn't complain about the status code.

computate:
Thanks for the suggestions @larsks , I have made those updates, please review again.

prom-keycloak-proxy/overlays/nerc-ocp-obs/deployments/prom-keycloak-proxy/deployment.yaml
larsks: Many applications provide an unauthenticated healthcheck endpoint for exactly this reason. Maybe that's something we want to do with the proxy? We can keep that in mind as a future change.

computate: Good suggestion @larsks , I will plan on adding a better health check.

computate:
Finally got the `envFrom` section right, now it's ready @larsks .

prom-keycloak-proxy/overlays/nerc-ocp-obs/deployments/prom-keycloak-proxy/deployment.yaml
```diff
@@ -0,0 +1,73 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: prom-keycloak-proxy
+  namespace: prom-keycloak-proxy
+spec:
+  replicas: 1
+  revisionHistoryLimit: 0
+  strategy:
+    type: Recreate
```
larsks: Why does this deployment need to be `type: Recreate` (typically we only see that with things like databases that require exclusive ownership of some filesystem resources)?

computate: @larsks It doesn't need to be Recreate, I switched it to RollingUpdate instead.

larsks (approved):

PR #464, opened 2024-07-10
nerc-ocp-infra: upgrade OCP 4.14.28 -> 4.15.16

dystewart:
There is no admin ack required for this upgrade, since no apis are removed


schwesig (approved):
:+1:

computate (approved):

PR #480, opened 2024-07-29
Create Kruize-admins cluster roles

tssala23:
Previously the kruise admins group was binded to the the nerc-ops cluster role. The PR creates a kruize-admins cluster role which aggregates roles from cluster-admin labels as opposed to cluster-reader. This will give the kruize admin group full admin access to the cluster without having to use the `--as system:admin` flag.

### DIFF
--- a/cluster-scope/base/rbac.authorization.k8s.io/clusterroles/kruize-admins/clusterrole.yaml
+++ b/cluster-scope/base/rbac.authorization.k8s.io/clusterroles/kruize-admins/clusterrole.yaml
@@ -1,0 +1,12 @@
+aggregationRule:
+  clusterRoleSelectors:
+    - matchLabels:
+        nerc.mghpcc.org/aggregate-to-nerc-ops: "true"
+    - matchLabels:
+        rbac.authorization.k8s.io/aggregate-to-admin: "true"
+    - matchLabels:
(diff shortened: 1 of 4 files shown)


cluster-scope/base/rbac.authorization.k8s.io/clusterroles/kruize-admins/clusterrole.yaml
```diff
@@ -0,0 +1,12 @@
+aggregationRule:
+  clusterRoleSelectors:
+    - matchLabels:
+        nerc.mghpcc.org/aggregate-to-nerc-ops: "true"
+    - matchLabels:
+        rbac.authorization.k8s.io/aggregate-to-cluster-admin: "true"
```
larsks: There is no existing role with a matching `aggregationRule` (so . I think you want:

```
    - matchLabels:
        rbac.authorization.k8s.io/aggregate-to-admin: "true"
```

Compare:

```
kubectl get clusterrole -l rbac.authorization.k8s.io/aggregate-to-cluster-admin=true
```

With:

```
kubectl get clusterrole -l rbac.authorization.k8s.io/aggregate-to-admin=true
```

larsks (approved):

dystewart (approved):

schwesig (commented):
💚

PR #484, opened 2024-08-01
Grant nerc-ops read permissions on RHOAI resources

larsks:
This grants us read permissions on resources in the
`datasciencepipelinesapplications.opendatahub.io` and `serving.kserve.io`
API groups, which will permit us to view items in the "Data Science
Projects" and "Model Serving" tabs of the RHOAI web interface.


computate (approved):

schwesig (approved):
Thanks @larsks @computate for taking care of this.

PR #490, opened 2024-08-09
Add nncp for gpu node

tssala23:
This node network configuration policy is needed for the GPU node to gain access to NESE storage. Unlike the other nodes on the cluster the GPU node only has a single interface, this meant that we created a trunk port in order to gain access to the private network that the node are on and the nese storage network.
Additional configuration is need on the node for this to work; create a vlan interface with a tag on the host and attach it to the trunk port.

### DIFF
--- a/cluster-scope/overlays/nerc-ocp-test-2/nodenetworkconfigurationpolicies/vlan-211.yaml
+++ b/cluster-scope/overlays/nerc-ocp-test-2/nodenetworkconfigurationpolicies/vlan-211.yaml
@@ -1,0 +1,34 @@
+apiVersion: nmstate.io/v1
+kind: NodeNetworkConfigurationPolicy
+metadata:
+  name: vlan-211
+spec:
+  nodeSelector:
+    kubernetes.io/hostname: "wrk-5"
+  desiredState:
+    interfaces:
+    - name: eno1np0.211
+      type: vlan
+      state: up
+      ipv4:
+        enabled: true
+        dhcp: false
+        auto-routes: false
+        address:
+        - ip: 10.0.120.39
+          prefix-length: 22
+      vlan:
+        base-iface: eno1np0
+        id: 211
+      mtu: 9000
+    routes:
+      config:
+      - destination: 10.255.116.0/23
+        next-hop-address: 10.0.120.1
+        next-hop-interface: vlan-211
+      - destination: 10.247.236.0/25
+        next-hop-address: 10.0.120.1
+        next-hop-interface: vlan-211
+      - destination: 140.247.236.0/25
+        next-hop-address: 10.0.120.1
+        next-hop-interface: vlan-211
--- a/cluster-scope/overlays/nerc-ocp-test-2/nodenetworkconfigurationpolicies/vlan-211.yaml
+++ b/cluster-scope/overlays/nerc-ocp-test-2/nodenetworkconfigurationpolicies/vlan-211.yaml
@@ -11,6 +11,9 @@ spec:
       type: vlan
       state: up
       ipv4:
+        # DHCP for this network is currently managed by Neutron, and Neutron is
+        # unable to provide DHCP service to trunked interfaces. We need to manually
+        # manage address assignment here until we move DHCP away from Neutron.
         enabled: true
         dhcp: false
         auto-routes: false
--- a/cluster-scope/overlays/nerc-ocp-test-2/nodenetworkconfigurationpolicies/kustomization.yaml
+++ b/cluster-scope/overlays/nerc-ocp-test-2/nodenetworkconfigurationpolicies/kustomization.yaml
@@ -2,3 +2,4 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 resources:
 - eth-nese.yaml
+- vlan-211.yaml


larsks (changes requested):
This should not be a VLAN interface. Therer are no other networks attached to this interface; we don't need a trunked VLAN.

We had access to NESE working  fine on the other machines without this configuration; we shouldn't need the VLAN interface now. The existing `eth-nese.yaml` should be sufficient.

tssala23:
@larsks yes but the gpu node only had one interface while the other node in the cluster had 2. With the other node I was able to connect them to the private network and the nese storage network with esi, however, with the gpu node we created a trunk port using esi to connect to both networks.
```
(.venv) [tsalawu@tsalawu-thinkpadx1nanogen2 mocesi]$ openstack esi node network list
+--------------------+-------------------+-------------------------------------+--------------------+----------------+------------------+-------------+
| Node               | MAC Address       | Port                                | Network            | Fixed IP       | Floating Network | Floating IP |
+--------------------+-------------------+-------------------------------------+--------------------+----------------+------------------+-------------+
| MOC-R4PAC22U31-S3A | e0:d8:48:e5:ae:5b | esi-MOC-R4PAC22U31-S3A-tsalawu      | tsalawu (448)      | 192.168.50.149 | None             | None        |
| MOC-R4PAC22U31-S3A | e0:d8:48:e5:ae:5e | esi-MOC-R4PAC22U31-S3A-nese-storage | nese-storage (211) | 10.0.123.163   | None             | None        |
| MOC-R8PAC23U39     | 08:8f:c3:a6:03:8e | esi-gpu-trunk-tsalawu-trunk-port    | tsalawu (448)      | 192.168.50.98  | None             | None        |
|                    |                   |                                     | nese-storage (211) | 10.0.122.169   |                  |             |
+--------------------+-------------------+-------------------------------------+--------------------+----------------+------------------+-------------+
```
I then ran these command on the node itself to get everything working
```
sh-5.1# nmcli connection add type vlan con-name vlan211 ifname vlan211 vlan.parent eno1np0 vlan.id 211
Connection 'vlan211' (e4d06285-2421-40d0-94bb-f21816dfb7ca) successfully added.
sh-5.1# nmcli connection modify vlan211 ipv4.addresses '10.0.120.39/22' ipv4.method manual
sh-5.1# nmcli connection modify vlan211 +ipv4.routes "10.255.116.0/23 10.0.120.1"
sh-5.1# nmcli connection modify vlan211 +ipv4.routes "10.247.236.0/25 10.0.120.1"
sh-5.1# nmcli connection modify vlan211 +ipv4.routes "140.247.236.0/25 10.0.120.1"
sh-5.1# nmcli connection up vlan211
```
Once that was done I was able to ping storage which was working prior.

larsks:
Ah, I didn't realize we had any single interface machines. Color me surprised!

larsks (dismissed):

cluster-scope/overlays/nerc-ocp-test-2/nodenetworkconfigurationpolicies/vlan-211.yaml
```diff
@@ -0,0 +1,34 @@
...
+        auto-routes: true
+        address:
+        - ip: 10.0.120.39
+          prefix-length: 22
+      vlan:
+        base-iface: eno1np0
+        id: 211
+      mtu: 9000
+    routes:
+      config:
```
larsks, line 28: Are you certain we need these static routes? This information should be provided by DHCP.

cluster-scope/overlays/nerc-ocp-test-2/nodenetworkconfigurationpolicies/vlan-211.yaml
```diff
@@ -0,0 +1,34 @@
...
+    interfaces:
+    - name: eno1np0.211
+      type: vlan
+      state: up
+      ipv4:
+        enabled: true
+        dhcp: true
+        auto-routes: true
+        address:
+        - ip: 10.0.120.39
```
larsks, line 21: Why are we setting a static ip here if we have `dhcp: true`?

schwesig (approved):
@tssala23 will this fix
- https://github.com/nerc-project/operations/issues/679

larsks (commented):
A couple of inline comments, but if the new machines only have a single interface, won't the existing `eth-nese.yaml` fail to apply? Will we need to update the `nodeSelector` to exclude the GPU node(s)?

cluster-scope/overlays/nerc-ocp-test-2/nodenetworkconfigurationpolicies/vlan-211.yaml
tssala23: "I don't think there's any dhcp for tagged networks" - [Mainn](https://massopencloud.slack.com/archives/CG1C4SZCM/p1723046737555839?thread_ts=1722974917.195019&cid=CG1C4SZCM) 
When the vlan interface was created it did not get assigned an ip, there's a discussion in the linked thread about it. Hakan said to set that static IP.

cluster-scope/overlays/nerc-ocp-test-2/nodenetworkconfigurationpolicies/vlan-211.yaml
tssala23: This should be set to `false` will make the change

cluster-scope/overlays/nerc-ocp-test-2/nodenetworkconfigurationpolicies/vlan-211.yaml
larsks: We should be able to use DHCP. Tagged/not tagged doesn't have any impact on whether or not there is DHCP available.

tssala23:
> A couple of inline comments, but if the new machines only have a single interface, won't the existing `eth-nese.yaml` fail to apply? Will we need to update the `nodeSelector` to exclude the GPU node(s)?

@larsks Yes the `eth-nese.yaml` will fail to apply to the GPU node so we will need to update the `nodeSelector` to exclude the GPU node. What would be the best way to change the `nodeSelector`, would it be to make separate labels for gpu node and non-gpu nodes? Currently this nncp is just selecting the node by host name.
I am not sure if that project ill be requiring anymore nodes added to that cluster.

cluster-scope/overlays/nerc-ocp-test-2/nodenetworkconfigurationpolicies/vlan-211.yaml
tssala23: Maybe leave a comment on the original thread or in the esi channel as I don't know why we can't use DHCP for this instance

tssala23: Discussed in the check in meeting

larsks:
I updated the PR to add a comment to the NNCP manifest about the DHCP situation.

larsks (approved):

PR #492, opened 2024-08-12
Update limitrange

naved001:
The default container size aligns with our SU definitions and now there's a minimum size for a container.

Related: https://github.com/CCI-MOC/openshift-acct-mgt/pull/102

It's not clear to me how the existing limitranges will be updated with the new values.

### DIFF
--- a/acct-mgt/base/configmaps.yaml
+++ b/acct-mgt/base/configmaps.yaml
@@ -54,14 +54,18 @@ data:
         {
             "type": "Container",
             "default": {
-                "cpu": "2",
-                "memory": "1024Mi",
+                "cpu": "1",
+                "memory": "4096Mi",
                 "nvidia.com/gpu": "0"
             },
             "defaultRequest": {
-                "cpu": "1",
-                "memory": "512Mi",
+                "cpu": "500m",
+                "memory": "2048Mi",
                 "nvidia.com/gpu": "0"
+            },
+            "min": {
+              "cpu": "1m",
+              "memory": "1Mi"
             }
         }
     ]


larsks:
> It's not clear to me how the existing limitranges will be updated with the new values.

Right, I'm not sure if we have a mechanism that updates existing projects to match the updated configuration. @knikolla?

larsks (approved):

knikolla:
> > It's not clear to me how the existing limitranges will be updated with the new values.
> 
> Right, I'm not sure if we have a mechanism that updates existing projects to match the updated configuration. @knikolla?

A script will be need to be written that lists all the namespaces and updates the limitrange values. Since they all have the same value it shouldn't be that hard.

larsks:
> A script will be need to be written that...

@knikolla should this be a controller of some sort that runs periodically on the managed clusters?

knikolla:
> > A script will be need to be written that...
> 
> 
> 
> @knikolla should this be a controller of some sort that runs periodically on the managed clusters?

@larsks no, this only needs to run once to apply the limits to already existing namespaces. New namespaces will be created with the correct limitrange.

naved001:
I am marking this as draft because I was looking at some pods in the rhods-notebook namespace and it looks like some of the containers are running with minimal resources: 

```
    resources:
      limits:
        cpu: 100m
        memory: 64Mi
      requests:
        cpu: 100m
        memory: 64Mi
   ```
   
   I don't want to break things by setting the minimum too high.

naved001:
After some discussion with @joachimweyl we decided to set the minimum container size to reaaally small. That way these projects will still have some non-zero bill and it will show up in reports without preventing legitimately tiny containers from running.

There's not much we can do if users run their pod with artificially low resource requests as long as their pods can burst up to limits. Ultimately, this will only be solved when the overall cluster utilization increases and users will need to request the "true" amount necessary for their pods and can no longer rely on limits.

larsks (approved):

naved001:
@jtriley Wanted to make sure you have had a chance to look at this.

jtriley (approved):

PR #496, opened 2024-08-19
Fix duration in ingress cert on beta cluster

tssala23:
Correct multiple issues with default-ingress-certificate

- This Certificate resource was created in the `openshift-config` namespace,
  but it needs to be created in the `openshift-ingress` namespace [1].

- This Certificate was created with a duration of `360h0m0s` (15 days), but
  it should have a duration of `2160h0m0s` (90 days).

[1]: https://docs.openshift.com/container-platform/4.8/security/certificates/replacing-default-ingress-certificate.html

Signed-off-by: tssala23 <tsalawu@redhat.com>
Signed-off-by: Lars Kellogg-Stedman <lars@redhat.com>

### DIFF
--- a/cluster-scope/overlays/nerc-ocp-beta-test/certificates/default-ingress-certificate.yaml
+++ b/cluster-scope/overlays/nerc-ocp-beta-test/certificates/default-ingress-certificate.yaml
@@ -2,13 +2,13 @@ apiVersion: cert-manager.io/v1
 kind: Certificate
 metadata:
   name: default-ingress-certificate
-  namespace: openshift-config
+  namespace: openshift-ingress
 spec:
   issuerRef:
     name: letsencrypt-production-dns01
     kind: Issuer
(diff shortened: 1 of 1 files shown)


dystewart (approved):

dystewart:
I'm curious if the cluster will complain about updating the cert, or if it'll need to be recreated

tssala23:
> I'm curious if the cluster will complain about updating the cert, or if it'll need to be recreated

I’ve already applied the correct cert to the cluster. It never created the cert due to that value being incorrect

larsks (changes requested):
It would be great to see more information in the commit message (what was the problem, what's 2160h in days, what the behavior was with the previous value, etc)

larsks:
I have updated this PR to (a) fix the namespace in which we're creating the default-ingress-certificate and (b) include more information in the commit message.

larsks (approved):

PR #507, opened 2024-08-26
Change nncp to match new label added to nodes

tssala23:
There are two types of worker nodes, non-GPUs and GPUs. The GPUs on this cluster only have one interface, this means that networking for these nodes are slightly different. This change makes it so this nncp is only applied to the non-GPU nodes.
Labels were added to the non-gpu nodes:
`kubectl label node <non-gpu-wrk> gpu=false --as system:admin`

### DIFF
--- a/cluster-scope/overlays/nerc-ocp-beta-test/nodenetworkconfigurationpolicies/eth-nese.yaml
+++ b/cluster-scope/overlays/nerc-ocp-beta-test/nodenetworkconfigurationpolicies/eth-nese.yaml
@@ -4,7 +4,7 @@ metadata:
   name: eth-nese
 spec:
   nodeSelector:
-    node-role.kubernetes.io/worker: ""
+    gpu: "false"
   desiredState:
     interfaces:
     - name: mlx1


larsks (commented):
Are we running the nvidia node discovery operator on this cluster? I think we should be if we have GPU nodes.

jtriley:
> Are we running the nvidia node discovery operator on this cluster? I think we should be if we have GPU nodes.

@tssala23 To get the nvidia-gpu and node-feature-discovery operators working on a cluster:

1. Include `node-feature-discovery` bundle in your cluster-scope overlay's top-level `kustomization.yaml`
1. Include `nvidia-gpu-operator` bundle in your cluster-scope overlay's top-level `kustomization.yaml`
1. Create a cluster overlay in the `nfd-discovery` top-level directory
1. Create a cluster overlay in the `nvidia-gpu-operator` top-level directory
1. Add a `node-feature-discovery` argocd app to to the `nerc-ocp-apps` repo
1. Add an `nvidia-gpu-operator` argocd app to the `nerc-ocp-apps` repo

For the argocd apps, see:

https://github.com/OCP-on-NERC/nerc-ocp-apps/blob/main/clusters/nerc-ocp-test/kustomization.yaml#L40-L54

joachimweyl:
@tssala23 what are the next steps on this?

tssala23:
@joachimweyl  This can be put in the icebox for now. We implemented the change without using nncp and the nncp dont currently work properly though I do want to come back to it at some point, Not important now.

PR #530, opened 2024-09-09
ocp-test: add missing kubevirt manifests

jtriley:
The default install of the kubevirt/cnv operator seems to be missing several manifests that should be in the default namespace. Adding them to the nerc-ocp-test overlay for now until we hear back from RH support on whether these should come out of the box or if we're expected to configure these ourselves.

See:

- https://github.com/redhat-cop/agnosticd/blob/development/ansible/roles_ocp_workloads/ocp4_workload_virt_roadshow_multi_user/tasks/workload.yml#L22
- https://github.com/redhat-cop/agnosticd/tree/development/ansible/roles_ocp_workloads/ocp4_workload_virt_roadshow_multi_user/templates

### DIFF
--- a/cluster-scope/base/core/configmaps/kubevirt/configmap-ui-settings.yaml
+++ b/cluster-scope/base/core/configmaps/kubevirt/configmap-ui-settings.yaml
@@ -1,0 +1,16 @@
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: kubevirt-ui-features
+  namespace: default
+data:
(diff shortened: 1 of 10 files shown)


jtriley:
This is live already on the cluster, fixes the install, and redhat-cop repo has an ansible task called `Set up user permissions to fix missing kubevirt permissions` to apply these manifests. We should go ahead and merge this and we can make changes if RH support has suggestions.

tssala23 (approved):

naved001 (approved):

PR #533, opened 2024-09-11
ocp-test: add udev rules for v100 worker nics

jtriley:
This is needed in order to properly configure the bond and tagged storage NICs on the V100 GPU worker node `wrk-3`

(I also verified these NICs do not exist on the other 3x worker nodes so there shouldn't be any changes on the other host aside from a reboot to pick up the new udev rules which will do nothing on those hosts)


computate (approved):

tssala23 (approved):
lgtm

PR #537, opened 2024-09-18
Fix up RBAC for matt brewster/project robbie

larsks:
This corrects a problem with the existing RBAC that was preventing code in
project robbie from performing the equivlant of:

    $ id -un
    positron
    $ su -c date job_user
    su: cannot set groups: Operation not permitted

The problem here was caused by the use of the `SYS_ADMIN` capability;
without that capability, everything worked as expected:

    $ su -c date job_user
    Wed Sep 18 20:36:25 UTC 2024

This happens because without the SYS_ADMIN capability, the pod would bind
to the `anyuid` scc, which doesn't explicitly drop anything other than the
MKNOD capability. When we add the `SYS_ADMIN` capability to the pod, this
causes it instead to bind to the `nerc-allow-sys-admin` scc, which only
allowed *particular* capabilities. This results in a pod that looks like
this:

    spec:
      containers:
      - securityContext:
        allowPrivilegeEscalation: true
        capabilities:
          add:
          - SYS_ADMIN
          drop:
          - ALL

That is, while the pod has the `SYS_ADMIN` capability, it drops everything
else, include the ability to use setuid/setgid binaries, which causes the
above error.

The solution is to replace the two separate sccs (`anyuid` +
`nerc-allow-sys-admin`) with a single scc that grants all the necessary
privileges. This commit drops the rolebinding permitting use of the
`anyuid` scc, and modifies the `nerc-allow-sys-admin` scc to also grant the
`SETUID` and `SETGID` capabilities along with permitting `RunAsAny`. With
these changes, the operation above succeeds even when the pod has the
`SYS_ADMIN` capability.


Milstein (approved):

jtriley (approved):
Just a note we'll need to manually clean up the anyuid rolebinding after this gets merged.

PR #538, opened 2024-09-19
add nfs-rwx manifests and bundle

jtriley:
This implements an NFS server deployment on OpenShift and provisioner in order to provide RWX volumes. 

From: https://github.com/kwkoo/openshift-nfs-server

See: https://github.com/nerc-project/operations/issues/737

PR #539, opened 2024-09-20
csi-driver-nfs + nfs-server manifests for RWX storage

jtriley:
Alternate solution for https://github.com/nerc-project/operations/issues/737 using csi-driver-nfs (https://github.com/openshift/csi-driver-nfs)

This approach appears to be more maintained upstream than the deployment used in https://github.com/OCP-on-NERC/nerc-ocp-config/pull/538

Related nerc-ocp-apps PR: https://github.com/OCP-on-NERC/nerc-ocp-apps/pull/55

### DIFF
--- a/csi-driver-nfs/overlays/nerc-ocp-test/kustomization.yaml
+++ b/csi-driver-nfs/overlays/nerc-ocp-test/kustomization.yaml
@@ -1,0 +1,7 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+commonLabels:
+  nerc.mghpcc.org/kustomized: "true"
+
+resources:
+- ../../base
--- a/csi-driver-nfs/base/core/daemonsets/csi-driver-node.yaml
+++ b/csi-driver-nfs/base/core/daemonsets/csi-driver-node.yaml
@@ -1,0 +1,143 @@
+---
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: csi-nfs-node
+  namespace: csi-driver-nfs
+spec:
+  selector:
+    matchLabels:
+      app: csi-nfs-node
+  template:
+    metadata:
+      labels:
+        app: csi-nfs-node
+    spec:
+      containers:
+      - args:
+        - --csi-address=/csi/csi.sock
+        - --probe-timeout=3s
+        - --http-endpoint=localhost:29653
+        - --v=2
+        image: registry.k8s.io/sig-storage/livenessprobe:v2.13.1
+        name: liveness-probe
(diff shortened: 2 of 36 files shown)


jtriley:
This is live on ocp-test and confirmed working. This is an improvement over https://github.com/OCP-on-NERC/nerc-ocp-config/pull/538

larsks (commented):
Can we move all the non-cluster-scoped manifests into a top-level directory and create a new app?

...or maybe everything, if splitting it seems weird. I'm flexible on that point; I'm just trying to get all the non-cluster-scoped stuff out of the "cluster-scoped" directory. This is effectively an app install, just like vault, dex, acm, etc.

tssala23:
@jtriley Thanks for getting this working

tssala23:
@larsks for the time being should I deploy the bundle to the cluster that needed this?

larsks:
@tssala23 If we can get this cleaned up today, I'd prefer to wait. If it looks like it's going to take longer, just run with it so the ET folks aren't blocked.

jtriley:
@larsks @tssala23 I've cleaned this up as requested. This will also need to be reviewed and merged:

https://github.com/OCP-on-NERC/nerc-ocp-apps/pull/55

csi-driver-nfs/overlays/nerc-ocp-test/kustomization.yaml
```diff
@@ -0,0 +1,16 @@
...
+- ../../base
+
+patches:
+- target:
+    kind: StorageClass
+    name: nfs-csi
+  patch: |
+    - op: replace
+      path: /parameters/server
+      value: nfs-server.nfs.svc.cluster.local
```
larsks: Since we're always going to be using the in-cluster address -- which will be the same on very cluster -- should we just move this into `../../base`?

larsks (approved):

tssala23 (approved):

csi-driver-nfs/overlays/nerc-ocp-test/kustomization.yaml
jtriley: I had it that way originally and changed it given that it's not necessarily a given that a cluster is going to use the local NFS deployment. For example, I'm looking into configuring NFS on the weka storage and using that with the csi-driver-nfs deployment on the prod cluster. That said, and as you mentioned, since a bunch of test clusters are going to use the locally-deployed NFS setup we can set the default to the locally deployed endpoint and patch as needed.

jtriley: I've updated the PR as requested.

PR #543, opened 2024-09-24
Add oauth-proxy imagestream

IsaiahStapleton:
This image is used by RHOAI to provide authentication. RHOAI is currently pulling this image externally but we would like to store it and pull from the internal image repo.

### DIFF
--- a/cluster-scope/overlays/nerc-ocp-prod/rhoai/imagestreams/oauth-proxy.yaml
+++ b/cluster-scope/overlays/nerc-ocp-prod/rhoai/imagestreams/oauth-proxy.yaml
@@ -1,0 +1,17 @@
+kind: ImageStream
+apiVersion: image.openshift.io/v1
+metadata:
+  name: oauth-proxy
+  namespace: redhat-ods-applications
+spec:
+  lookupPolicy:
+    local: false
+  tags:
+    - name: latest
+      from:
+        kind: DockerImage
+        name: 'registry.redhat.io/openshift4/ose-oauth-proxy@sha256:4bef31eb993feb6f1096b51b4876c65a6fb1f4401fee97fa4f4542b6b7c9bc46'
+      importPolicy:
+        scheduled: true
+      referencePolicy:
+        type: Source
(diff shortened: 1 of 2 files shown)


computate:
@IsaiahStapleton Don't forget to enable the pre-commit-checks so you don't miss the yaml validation issues. 
https://github.com/OCP-on-NERC/nerc-ocp-config?tab=readme-ov-file#pre-commit-checks

IsaiahStapleton:
@computate Fixed, thanks

computate (commented):
@IsaiahStapleton Why are we introducing the exact image that already exists? 

```bash
$ oc -n redhat-ods-applications get imagestream/oauth-proxy -o yaml | grep from: -A 2
    from:
      kind: DockerImage
      name: registry.redhat.io/openshift4/ose-oauth-proxy@sha256:4bef31eb993feb6f1096b51b4876c65a6fb1f4401fee97fa4f4542b6b7c9bc46
```

It doesn't seem like a good idea to hard-code the image with this sha256. I imagine OpenShift automatically updates this over time. Maybe I don't understand the goals of this. Maybe you could link it to an issue.

DanNiESh (approved):

IsaiahStapleton:
@computate The image already exists in redhat-ods-applications because I manually applied it there. I created this PR so we have a record of it and so ArgoCD can manage it (in the case resources get deleted). The reason I used the exact sha256 is because that is the exact one RHOAI uses but it pulls from an external registry. Last semester when Red Hat's registry was down students couldn't start pods because RHOAI was trying to pull this oauth-proxy container from the external registry rather than the internal one. So the idea was to store the image internally and then use a Gatekeeper policy to change where the pods are pulling the oauth-proxy container from. The relevant issue is here: https://github.com/nerc-project/operations/issues/506.

larsks (changes requested):
This needs to be included in a kustomization file.

computate (approved):

larsks (approved):

PR #550, opened 2024-10-01
clusterissuer: add aws-route53-credentials externalsecret

jtriley:
This was missing from the `nerc-certificate-clusterissuers` bundle so the manually created aws-route53-credentials secret that we created on the nerc-ocp-obs cluster during bootstrapping wasn't being synced with vault.

See: https://github.com/nerc-project/operations/issues/750


computate (approved):

tssala23 (approved):

jtriley:
This is likely in the wrong spot. The http01 challenge doesn't require route53 credentials - the DNS01 challenge does.

jtriley:
I updated this to move the externalsecret to the `nerc-certificate-clusterissuers` bundle instead.

tssala23 (approved):

computate (approved):

PR #556, opened 2024-10-03
fix kind in clusterissuer-dns01 externalsecret

jtriley:
Minor fix to #554 to update kind to ClusterSecretStore in the secreStoreRef. (I need more coffee 😛 )


jtriley:
I manually applied this on the obs cluster and confirmed it's the last "minor fix" needed for this secret 🙃

computate (approved):

PR #563, opened 2024-10-08
Enable PCI passthrough on nerc-ocp-test cluster

naved001:
This PR does 3 things:

* Enable IOMMU
* Bind device (the GPU, V100) to the vfio driver
* permit device passthrough in the kubevirt resource

Before merging, I need to apply a label to the node which feels like would be bunch of work with argocd.
```
oc label node wrk-3 nvidia.com/gpu.deploy.operands=false 
```

It would be interesting to see how this would work with the A100 systems considering each host has multiple identical GPUs and they will all have the same vendor and device PCI IDs. 

I am going to mark it as draft because the GPU is in use for other testing.

https://docs.openshift.com/container-platform/4.15/virt/virtual_machines/advanced_vm_management/virt-configuring-pci-passthrough.html


schwesig (approved):
besides waiting for free GPUs and the manual labels,
👍 LGTM

naved001:
@computate This is ready for review. Is it okay if I apply the label to wrk-3 which will kick off some nvidia pods from it? (cc @dystewart )

computate (approved):

computate:
Fine with me @naved001 , nice work here!

larsks (approved):
That all looks reasonable.

PR #570, opened 2024-10-17
ocp-test: add support for local users via htpasswd

jtriley:
This adds an htpasswd component that includes a patch to the OAuth resource and an externalsecret for the htpasswd file contents. This will enable us to create local user accounts for testing as non-admins or other use cases (e.g. monitoring via nagios for example)


jtriley:
Should also be useful toward https://github.com/nerc-project/operations/issues/766

larsks (approved):
Looks good. Thanks for implementing this as a kustomize component so we can potentially use it elsewhere!

larsks:
I added some detail to the first commit message in this PR.

computate (approved):

aabaris (approved):

PR #579, opened 2024-10-22
ocp-test: add udev rules for a100 mlx5 nics

jtriley:
This adds udev rules for the Mellanox nics connected to the A100 host added via ESI (wrk-4) in the nerc-admins project (MOC-R8PAC23U27). These rules were manually put in place and verified working.


tssala23 (approved):
lgtm thank you

computate (approved):

PR #590, opened 2024-11-04
Adding postgresql-ha on obs cluster

RH-csaggin:
- created all the postgresql-ha config resources for the ai-telemetry project
- creation of external secrets
- creation of namespace


RH-csaggin:
need to update it with right resources request and limts

PR #594, opened 2024-11-07
feat: nerc-obs switching grafana to v1beta1

schwesig:
# feat: nerc-obs switching grafana to v1beta1


RH-csaggin (approved):

schwesig:
change of plans,
we will add another grafana instance, to avoid the rewriting, changing fields and mitigating the many changes between the two versions, to not block the grafana dashboards we are currently using

PR #605, opened 2024-11-21
disable automatic NVIDIA driver upgrades

jtriley:
Automatic upgrades of the NVIDIA driver are dangerous in production given that they often happen unattended and are likely to cause issues on at least a subset of nodes. In the best case, user workloads are impacted at unpredictable times.

This patch disables automatic upgrades by default. Systems that want automatic upgrades should patch the GPU cluster policy in an overlay.


computate (approved):

tssala23 (approved):

schwesig:
thanks!!

PR #609, opened 2024-11-22
odf: enable ceph tools by default

jtriley:
This enables the ceph tools pod deployment in the openshift-storage by default given that it's useful for debugging issues and inspecting the current state of ceph.


tssala23 (approved):

computate (approved):

larsks:
Looks like we need to update the schema repository for this to validate correctly.

larsks (approved):

PR #610, opened 2024-11-26
Adding minio with dex auth to nerc-ocp-test cluster

computate:
Adding support for object stores in the test cluster with authentication
and policy based authorization by OpenShift groups.

### DIFF
--- a/minio/base/externalsecret-minio-admin-credentials.yaml
+++ b/minio/base/externalsecret-minio-admin-credentials.yaml
@@ -1,0 +1,15 @@
+apiVersion: external-secrets.io/v1beta1
+kind: ExternalSecret
+metadata:
+  name: minio-admin-credentials
+  namespace: minio
+spec:
+  refreshInterval: "1h"
+  secretStoreRef:
+    name: nerc-cluster-secrets
+    kind: ClusterSecretStore
+  target:
+    name: minio-admin-credentials
+  dataFrom:
+    - extract:
+        key: $ENV/$CLUSTER/minio/minio-config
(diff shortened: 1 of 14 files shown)


tssala23 (approved):

minio/base/externalsecret-minio-admin-credentials.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: external-secrets.io/v1beta1
+kind: ExternalSecret
+metadata:
+  name: minio-admin-credentials
+  namespace: minio
+spec:
+  refreshInterval: 15s
```
larsks: This seems like an aggressive refresh interval.

IsaiahStapleton (approved):

minio/base/kustomization.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+namespace: minio
+commonLabels:
+  app: minio
+  component: minio
```
larsks: You can drop this `component` label since you're not managing dex from the same kustomization file like I was in the demo repository.

minio/base/service.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: minio
+spec:
+  selector:
+    app: minio
```
larsks: Remove this; let kustomize manage selectors for you.

minio/base/externalsecret-minio-admin-credentials.yaml
computate: Sounds good, I made this change.

minio/base/kustomization.yaml
computate: Sounds good, I made this change.

minio/base/service.yaml
computate: Sounds good, I made this change.

PR #613, opened 2024-11-27
Fixing SecretStore reference for dex and minio on test

computate:
There isn't a ClusterSecretStore in the nerc-ocp-test cluster, so we
will use namespace SecretStores instead for dex and minio.


schwesig (approved):

tssala23 (approved):
The minio namespace will have to be added to the vault config for the test cluster, https://github.com/OCP-on-NERC/nerc-ocp-config/blob/main/vault/config/overlays/nerc-ocp-infra/config/nerc-ocp-test.yaml

computate:
Thanks for catching that @tssala23, it's now there.

PR #615, opened 2024-11-27
Fix some issues with virt bundle

larsks:
- We are installing  a subscription for the `kubevirt-hyperconverged`
  operator, but the subscription was named `hco-operatorhub`.

- The operatorgroup had an unnecessary `-group` suffix, which doesn't match
  our other installs.

- The bundle included resources which, according to cbea92581f, it's not
  clear if we need or not. Move this into the test cluster overlay for now
  in line with the comment in that commit.


computate (approved):

larsks:
@computate Because we're renaming a subscription here, this may require manual intervention on the test cluster when it merges. I will ping you before that happens.

tssala23 (approved):

joachimweyl:
@larsks is this pr ready for merge?

joachimweyl:
On hold until we have a test cluster we can apply it to that is not being worked on.

tssala23 (approved):

PR #620, opened 2024-12-03
Explicitly create service account token for eso-vault-auth-token

larsks:
With newer versions of kubernetes/openshift, it is necessary to explicitly
request a long-lived token for service accounts [1].

[1]: https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#manually-create-a-long-lived-api-token-for-a-serviceaccount

### DIFF
--- a/cluster-scope/base/core/secrets/eso-vault-auth-token/kustomization.yaml
+++ b/cluster-scope/base/core/secrets/eso-vault-auth-token/kustomization.yaml
@@ -1,0 +1,5 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+namespace: external-secrets-operator
+resources:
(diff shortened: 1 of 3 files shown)


naved001:
@larsks how was the sa token created before this?

schwesig (approved):

larsks:
@naved001 prior to Kubernetes 1.24, when you create a serviceaccount, kubernetes would automatically create a matching Secret containing a long-lived service account token. For 1.24 and later, this secret is not created automatically.

It looks like OpenShift has adopted the same behavior with 4.16 and later.

For these newer releases, if you want a long lived token you have to explicitly request one. If you just need a temporary token, you should use the `kubectl create token` command.

naved001 (approved):

PR #628, opened 2024-12-16
Grant nerc-test-people reader access to nerc-test-cluster

larsks:
Update the oauth configuration to permit access by the nerc-test-people
group, and grant this group cluster-reader access.


DanNiESh (approved):

schwesig (approved):
💚

PR #631, opened 2025-01-06
Add custom dns manifests needed to connect to infra

tssala23:
For cluster created using ESI there are some additional steps needed to connect the to the infra cluster due to our current network step up.
Adding the nerc-infra-routed network to each node is the first step.
The second is to set up custom dns on the cluster using these manifests to resolve the host name


tssala23:
It would also require a patch to the dns operator that would look like this:
```
dns-patch.yaml

spec:
  servers:
  - name: mydomain
    zones:
    - mydomain.org
    forwardPlugin:
      policy: Random
      upstreams:
      - <cluster-ip>:8053 #oc -n custom-dns get svc --no-headers | awk '{ print $3 }'
```
However I am not sure exactly where that would go

larsks:
~You can use the service name instead of an ip address for specifying the custom server, like this:~


That was a lie. Sorry for the confusion.

larsks (approved):

custom-dns/base/deployment.yaml
```diff
@@ -0,0 +1,39 @@
...
+  replicas: 1
+  selector:
+    matchLabels:
+      app: dns
+      deployment: dns
+  template:
+    metadata:
+      labels:
+        app: dns
+        deployment: dns
```
larsks: All the labels should really be managed by kustomize, rather than hardcoded here.

tssala23:
@larsks Does this look correct, I removed the labels and added the patch to a file included in the kustomization

larsks:
@tssala23 this looks fine. Regarding patching the DNS operator, it's important to note that argocd doesn't explicitly support patching existing resources...but because it effectively uses `kubectl apply` to push the manifests into the cluster, it should generally do the right thing. In some cases, we makes things explicit by taking a copy of the existing (unpatched) resource and placing that in the config repository (and then patching it in an overlay), but I don't think we've been entirely consistent in how we manage this.

I would consider adding `port` definitions in your Deployment, so that you can then refer to the target port by name in your Service.

tssala23:
@larsks does that look right with the port definitions?

larsks:
Looks good to me!

computate (approved):

IsaiahStapleton (approved):

schwesig (approved):

PR #634, opened 2025-01-08
feat: add "managementState" field to OTelCollector to fix ArgoCD sync issue

dheerajodha:


larsks (commented):
In the future, something like this should be two different pull requests (because the two different changes aren't really related). 

It would also be helpful if you were to include some information about why the  lack of a `managementState` setting was causing issues with argocd.

computate:
Here is the ArgoCD error: 
image

error validating data: ValidationError(OpenTelemetryCollector.spec): missing required field "managementState" in io.opentelemetry.v1alpha1.OpenTelemetryCollector.spec

dheerajodha:
Thanks for the review @larsks!
I've updated the PR to do only one thing, and added the details related to the change, within the [commit message](https://github.com/OCP-on-NERC/nerc-ocp-config/pull/634/commits/7a4ae0470fb8bdd72e8f55503c684a2cad826310).

computate (approved):

larsks (approved):

PR #635, opened 2025-01-10
Add vault config for albany

tssala23:
Add vault config for albany cluster per support engineer suggesting this may be effect ODF upgrade


schwesig (approved):

larsks (approved):
Looks good, but note from the [README](https://github.com/OCP-on-NERC/nerc-ocp-config/tree/main/vault/config/overlays/nerc-ocp-infra#cluster-token-reviewer-tokens) that you will also need to manually add a token to the `vault-credentials` secret.

PR #640, opened 2025-01-22
Remove ai-telemetry client permissions to metrics

computate:
We should remove the ai-telemetry client permissions to metrics all
clusters and all namespaces. This will allow groups of users to be
granted access to metrics with different permissions instead of all
users having access as we build the new ai-telemetry application.


tssala23 (approved):

schwesig (approved):
👍

PR #644, opened 2025-01-30
ocp-test: add ODF daemonset tolerations for GPU nodes

jtriley:
See nerc-project/operations#903


jtriley:
Applied this manually and it fixes the issue.

PR #646, opened 2025-01-31
Upgrading autopilot to version v2.1.0

computate:
This is for the latest GPU health checks from IBM.

 Also allows autopilot containers to run privileged for nvidia-smi. Because of our use of ACCEPT_NVIDIA_VISIBLE_DEVICES_ENVVAR_WHEN_UNPRIVILEGED=false, we set privileged=true for the autopilot service account so it can load the NVIDIA tools like nvidia-smi to check on the GPU health, without actually claiming the GPU.


IsaiahStapleton (approved):

tssala23 (approved):

dystewart (approved):

computate:
Latest autopilot metrics and dashboards have been applied! 
image

PR #671, opened 2025-03-12
configures the H100 systems to be in PCI passthrough mode

naved001:
It applies the config to the custom machineconfig pool 4xlag which only includes the H100 systems.

I named the H100 resource like how the A100 was named.

Just so that we can see I got the right PCI ID:

```
sh-5.1# lspci -nnv | grep -i nvidia
lspci: Unable to load libkmod resources: error -2
  DeviceName: Nvidia 4-chip VPI CX7 - 1
06:00.0 3D controller [0302]: NVIDIA Corporation GH100 [H100 SXM5 80GB] [10de:2330] (rev a1)
  Subsystem: NVIDIA Corporation Device [10de:16c0]
  Kernel driver in use: nvidia
  DeviceName: Nvidia 4-chip VPI CX7 - 2
  ```
  
  I will apply the label to kick out the nvidia pods once this is approved.

### CODE REVIEW

**Review by computate — APPROVED** [review]

File: cluster-scope/overlays/nerc-ocp-test/machineconfigs/pci-passthrough/100-4xlag-iommu.yaml [hunk 1]
```diff
@@ -0,0 +1,12 @@
+apiVersion: machineconfiguration.openshift.io/v1
+kind: MachineConfig
+metadata:
+  labels:
+    machineconfiguration.openshift.io/role: 4xlag
```
jtriley: Might be better to just use the worker role instead of `4xlag` for now. That pool is a stop-gap solution to configure LACP with 4x NICs vs 2x NICs (default). I'm still experimenting with ways to avoid having to create a custom MCP and also haven't made a PR for that piece yet (or the udev rules which are manually in place for now).

naved001: @jtriley sure, I have updated this PR to use the `worker` role instead.

jtriley (approved):

PR #682, opened 2025-04-14
Add Hypershift2 overlay

tssala23:
This adds the cluster-scope overlay for the hypershift2 cluster, our
our experimental instance of hosted control planes [1].

[1]: https://www.redhat.com/en/topics/containers/what-are-hosted-control-planes


larsks (approved):
Looks good to me after a cursory glance. The proof will be in the deployment :).  You probably want to tag a couple additional reviewers.

computate (approved):

dystewart (approved):

PR #683, opened 2025-04-17
Add overlay for dublin test cluster

dystewart:
Overlay is for this cluster: [1033](https://github.com/nerc-project/operations/issues/1033)


tssala23:
I have not been adding test cluster overlays to the config repo any more as we do not sync them up with argoCD and they will need to be removed at some point anyway

dystewart:
@tssala23 This is true of test clusters and in general I agree, however I wanted to track things for this cluster overlay specifically since it can/will serve as a template for installing and configuring ibm scale storage (which has proven tricky) which we'll be using for ROCE/RDMA testing. Basically, it's easier to keep track of all the ibm scale storage config against an overlay. 

More so I think this is a way to force myself to document commit messages wrt ibm scale more verbosely than I would on a local branch lol. I'll leave as a draft for now anyhow

schwesig (approved):
...

PR #704, opened 2025-06-06
ocp-obs: add ipv4 forwarding machineconfig

jtriley:
This enables ipv4 forwarding on the OBS cluster. I've also moved this machineconfig into base given that it's used by a bunch of overlays currently. It likely should just go in common and be removed from individual overlays given that we need it pretty much everywhere.


computate (approved):

tssala23 (approved):

schwesig (approved):

jtriley:
closed in favor of #706

PR #721, opened 2025-06-16
Add Ibm scale operator and deployment resources

dystewart:
Adds ibm-scale operator and deployment resources. This configuration is the one we've used in the dublin cluster to connect to CNSA and test single interface I/O on the H100 nodes. 

This PR does not add the resources to any cluster overlay, it just codifies the deployment for future deployments.


tssala23:
@dystewart @larsks 
I thought we had decided to create a separate repo for this as I will likely get removed one day?

PR #726, opened 2025-07-24
Edu cluster overlay

tssala23:
Created with my templating playbook and then modified to reflect prod
There are some comments for things that may have to be changed or that I updated
I made a couple assumptions in this:
1. the address will be edu.nerc.mghpcc.org
2. the vault secret path will be nerc-ocp-edu

Keycloack and github id need to be added
Machine configs need to be added
nncp need to be added
The oauth dir seems to be redundant


jtriley:
Just noting I'm making a copy of this branch that lives in the nerc-ocp-config repo so we can collaborate. Also I'm pretty sure a nerc-ocp-infra upgrade commit snuck into this branch so I'll remove that as well.

jtriley:
closing in favor of #727

PR #729, opened 2025-08-01
Separate prom-keycloak-proxy instances for moc, innabox-dev, open-sovereign-ai-cloud

computate:
Deploying separate prom-keycloak-proxy instances for both moc and
innabox-dev ACM Hub metrics. The Keycloak client ID and secret are also
distinct between them, as well as the permissions granted to each client
service account.


schwesig (approved):

prom-keycloak-proxy/overlays/nerc-ocp-obs/innabox-dev/patches/deployment.yaml
```diff
@@ -0,0 +1,12 @@
+- op: replace
+  path: /metadata/name
```
larsks: This (and other patches to `metadata.name`) might be better handled using `nameSuffix: -innabox-dev` in your `kustomization.yaml`.

prom-keycloak-proxy/overlays/nerc-ocp-obs/innabox-dev/patches/deployment.yaml
```diff
@@ -0,0 +1,12 @@
...
+  value: prom-keycloak-proxy-innabox-dev
+- op: replace
+  path: /spec/template/spec/containers/0/env/0/value
+  value: innabox-dev
+- op: replace
+  path: /spec/template/spec/containers/0/envFrom/0/secretRef/name
+  value: prom-keycloak-proxy-innabox-dev-secrets
```
larsks: These patches are fragile because they depend on the ordering of your environment variables. If someone updates the deployment in `base` so that it looks like:

```
    spec:
      containers:
        - name: prom-keycloak-proxy
          envFrom:
            - secretRef:
                name: prom-keycloak-proxy-secrets
          env:
            - name: someOtherVariable
              value: foo
            - name: PROXY_ACM_HUB
              value: REPLACE_IN_OVERLAY
```

You end up with:

```
    spec:
      containers:
      - env:
        - name: someOtherVariable
          value: innabox-dev
        - name: PROXY_ACM_HUB
          value: REPLACE_IN_OVERLAY
```

A better solution is to use a strategic merge patch:

```
  - patch: |
      apiVersion: apps/v1
      kind: Deployment
      metadata:
        name: prom-keycloak-proxy-innabox-dev
        namespace: prom-keycloak-proxy
      spec:
        template:
          spec:
            containers:
            - env:
              - name: PROXY_ACM_HUB
                value: innabox-dev
              name: prom-keycloak-proxy
```

This will patch the correct variable regardless of its position (the same holds for the subsequent `envFrom` patch).

prom-keycloak-proxy/overlays/nerc-ocp-obs/innabox-dev/patches/deployment.yaml
computate: Thanks for the suggestions @larsks , I have applied your suggestions and updated my commit.

prom-keycloak-proxy/overlays/nerc-ocp-obs/innabox-dev/patches/deployment.yaml
computate: Thanks for the suggestions @larsks , I have applied your suggestions and updated my commit.

larsks (approved):

PR #732, opened 2025-08-07
Bind nerc-infra-routed nncp to worker nodes

larsks:
Some networking fixes necessary for the hypershift2 cluster.

The `ipForwarding: Global` change we probably want to move out of the hypershift2 overlay at some point and make it the default on our other clusters.

### DIFF
--- a/cluster-scope/overlays/hypershift2/kustomization.yaml
+++ b/cluster-scope/overlays/hypershift2/kustomization.yaml
@@ -127,3 +127,12 @@ patches:
       path: /spec/routeAdmission
       value:
         wildcardPolicy: WildcardsAllowed
+
+- target:
+    kind: Network
+    group: operator.openshift.io
+    name: cluster
+  patch: |
+    - op: add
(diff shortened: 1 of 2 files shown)


cluster-scope/base/operator.openshift.io/networks/cluster/network.yaml
```diff
@@ -7,3 +7,4 @@ spec:
     ovnKubernetesConfig:
       gatewayConfig:
         routingViaHost: true
+        ipForwarding: Global
```
jtriley: Did you mean to do this in base? I think `infra` is still on SDN networking

jtriley (dismissed):

cluster-scope/base/operator.openshift.io/networks/cluster/network.yaml
larsks: I did not! Good catch.

tssala23 (approved):

computate (approved):

PR #734, opened 2025-08-07
Add argocd post-sync hook to patch dns operator

larsks:
This adds a job that discovers the address of the custom dns service and
patches the dns operator appropriately. This will run automatically
whenever argocd pushes changes to the cluster.

The job will also run when you `kubectl apply` these manifests, but in
this case you will need to manually remove the job first if it has
previously been created (and if you are pushing a modified version).


tssala23 (approved):

tssala23 (approved):

jtriley:
@larsks @tssala23 Now that this is no longer needed with the rebuild of the infra cluster, do we want to merge this (for historical/posterity purposes) and then clean up the custom-dns manifests in another PR or just leave it around in case a similar use-case arises in the future?

tssala23:
I'm fine with either merging and leaving it or merging then removing it as you said for for historical/posterity purposes. I cannot think of a reason as to why we would need it in the future, but there could always be the odd case of maybe a test cluster we set up for someone needing to reach something maybe ¯\_(ツ)_/¯ no harm in keeping it around

larsks:
@jtriley I've rebased it so it can be merged. I'm happy either way.

tssala23 (approved):

computate (approved):

PR #748, opened 2025-08-13
Add gpu-taint-management bundle and resources

dystewart:
Addresses issue: https://github.com/nerc-project/operations/issues/1190

Added bundle to nerc-ocp-test cluster deployment for testing.

### DIFF
--- a/cluster-scope/bundles/gpu-taint-management/taint.sh
+++ b/cluster-scope/bundles/gpu-taint-management/taint.sh
@@ -1,0 +1,45 @@
+#!/bin/bash
+
+# GPU Taint Watcher Script
+
+# Initial tainting of GPU nodes
+kubectl get nodes -l node-role.kubernetes.io/worker -o json |
+jq -r '
+  .items[]
+  | select(.metadata.labels["nvidia.com/gpu.product"] != null)
+  | .metadata.name + " " + .metadata.labels["nvidia.com/gpu.product"]
+' | while read NODE PRODUCT; do
+  echo "Initial tainting: $NODE with GPU product $PRODUCT"
+  kubectl taint node "$NODE" "nvidia.com/gpu.product=$PRODUCT:NoSchedule" --overwrite=false 2>/dev/null
+done
+
+# Declare associative array to track last seen GPU product per node
+declare -A LAST_PRODUCT
+
+# Continuous monitoring loop
+while true; do
+  kubectl get nodes -l node-role.kubernetes.io/worker -o json |
+  jq -r '.items[].metadata.name' | while read NODE_NAME; do
+    PRODUCT=$(kubectl get node "$NODE_NAME" -o jsonpath="{.metadata.labels['nvidia\.com/gpu\.product']}" 2>/dev/null)
+
+    # Skip nodes without GPU product label
+    if [ -z "$PRODUCT" ]; then
+      echo "No GPU product label on $NODE_NAME, skipping"
+      continue
+    fi
+
+    # Check current taints on the node
+    CURRENT_TAINT=$(kubectl get node "$NODE_NAME" -o jsonpath="{.spec.taints[?(@.key=='nvidia.com/gpu.product')].value}" 2>/dev/null)
+
+    # Apply taint if label changed or taint is missing
+    if [[ "${LAST_PRODUCT[$NODE_NAME]}" != "$PRODUCT" || "$CURRENT_TAINT" != "$PRODUCT" ]]; then
+      echo "Applying taint for GPU product '$PRODUCT' on $NODE_NAME (current taint: '${CURRENT_TAINT:-none}')"
+      kubectl taint node "$NODE_NAME" "nvidia.com/gpu.product=$PRODUCT:NoSchedule" --overwrite=false 2>/dev/null
+      LAST_PRODUCT[$NODE_NAME]="$PRODUCT"
+    else
+      echo "No change for $NODE_NAME (product: $PRODUCT, taint present)"
+    fi
+  done
+
+  sleep 30
+done
--- a/cluster-scope/bundles/gpu-taint-management/clusterrole.yaml
+++ b/cluster-scope/bundles/gpu-taint-management/clusterrole.yaml
@@ -1,0 +1,8 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: gpu-product-taint-role
+rules:
+  - apiGroups: [""]
+    resources: ["nodes"]
+    verbs: ["get", "list", "patch", "watch"]
--- a/cluster-scope/bundles/gpu-taint-management/clusterrolebinding.yaml
+++ b/cluster-scope/bundles/gpu-taint-management/clusterrolebinding.yaml
@@ -1,0 +1,12 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: gpu-product-taint-rolebinding
+subjects:
+  - kind: ServiceAccount
+    name: gpu-product-taint-sa
+    namespace: gpu-taint-management
+roleRef:
+  kind: ClusterRole
+  name: gpu-product-taint-role
+  apiGroup: rbac.authorization.k8s.io
--- a/cluster-scope/bundles/gpu-taint-management/deployment.yaml
+++ b/cluster-scope/bundles/gpu-taint-management/deployment.yaml
@@ -1,0 +1,38 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: gpu-product-taint
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: gpu-product-taint
+  template:
+    metadata:
+      labels:
+        app: gpu-product-taint
+    spec:
+      serviceAccountName: gpu-product-taint-sa
+      affinity:
+        podAntiAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            - labelSelector:
+                matchExpressions:
+                  - key: app
+                    operator: In
+                    values:
+                      - gpu-product-taint
+              topologyKey: kubernetes.io/hostname
+      containers:
+      - name: gpu-taint
+        image: image-registry.openshift-image-registry.svc:5000/openshift/tools:latest
+        command: ["/bin/bash", "/script/taint.sh"]
+        volumeMounts:
+          - name: script-volume
+            mountPath: /script
+      restartPolicy: Always
+      volumes:
+        - name: script-volume
+          configMap:
+            name: gpu-product-taint-script
+            defaultMode: 0755
--- a/cluster-scope/bundles/gpu-taint-management/kustomization.yaml
+++ b/cluster-scope/bundles/gpu-taint-management/kustomization.yaml
@@ -1,0 +1,18 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+namespace: gpu-taint-management
+
+resources:
+  - deployment.yaml
+  - sa.yaml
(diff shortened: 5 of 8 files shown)


gpu-taint-management/base/daemonset.yaml
```diff
@@ -0,0 +1,48 @@
...
+          projected:
+            sources:
+              - serviceAccountToken:
+                  path: token
+                  expirationSeconds: 3600
+              - configMap:
+                  name: kube-root-ca.crt
+                  items:
+                    - key: ca.crt
+                      path: ca.crt
```
larsks: This does not need to be here (it is generated dynamically by kubernetes).

gpu-taint-management/base/configmap.yaml
```diff
@@ -0,0 +1,18 @@
...
+      echo "Checking node: $NODE_NAME"
+      PRODUCT=$(kubectl get node "$NODE_NAME" -o jsonpath="{.metadata.labels['nvidia\.com/gpu\.product']}")
+      if [ -n "$PRODUCT" ]; then
+        echo "Found GPU product label: $PRODUCT on $NODE_NAME"
+        kubectl taint node "$NODE_NAME" "nvidia.com/gpu.product=$PRODUCT:NoSchedule"
+      else
+        echo "No GPU product label found on node $NODE_NAME, skipping tainting"
+      fi
+      sleep 3600
+    done
```
larsks: Just an opinion, not a request for a change:

I like to put scripts in separate files and the build the configmap using a `configMapGenerator`. By having the script in a file by itself, rather than embedded in a YAML document, your editor/IDE can apply appropriate syntax checking and formatting.

```
configMapGenerator:
- name: gpu-product-taint-script
  options:
    disableNameSuffixHash: true
  files:
  - files/taint.sh
```

larsks: I wonder if instead of using a Daemonset with a script running for each node if it would make more sense to have a single instance using a kubernetes watch, something like:

```
#!/bin/bash

while true; do

  # watch for new/modified nodes
  kubectl get node -l node-role.kubernetes.io/worker -o name -w | while read NODE_NAME; do
    echo "Checking node: $NODE_NAME"
    PRODUCT=$(kubectl get "$NODE_NAME" -o jsonpath="{.metadata.labels['nvidia\.com/gpu\.product']}")
    if [ -n "$PRODUCT" ]; then
      echo "Found GPU product label: $PRODUCT on $NODE_NAME"
      kubectl taint "$NODE_NAME" "nvidia.com/gpu.product=$PRODUCT:NoSchedule"
    else
      echo "No GPU product label found on node $NODE_NAME, skipping tainting"
    fi
  done

  # watch expired, pause before restarting
  sleep 10
done
```

This will trigger whenever a node changes, so it will respond immediately either to new or modified nodes.

dystewart: @larsks I think this is a better approach, it ensures we aren't having any awkward time between when the new nodes are tainted and when the daemonset has the opportunity to taint the node. Plus it's a single instance

larsks: Or slightly different, moving the label check into the `kubectl get` command itself:

```
#!/bin/bash

while true; do

  # watch for new/modified nodes that have the nvidia.com/product label
  kubectl get node -w -l nvidia.com/gpu.product \
      --no-headers -o custom-columns='NAME:.metadata.name,PRODUCT:.metadata.labels.nvidia\.com/gpu\.product' |
    while read NODE_NAME PRODUCT; do
      current_taint_value=$(kubectl get node "$NODE_NAME" -o jsonpath='{.spec.taints[?(.key == "nvidia.com/gpu.product")].value}')
      desired_taint="nvidia.com/gpu.product=$PRODUCT:NoSchedule"
      if [[ $current_taint_value != "$PRODUCT" ]]; then
        echo "applying taint $desired_taint to node $NODE_NAME"
        kubectl taint node "$NODE_NAME" "$desired_taint" --overwrite --as system:admin
      else
        echo "node $NODE_NAME already has taint for $desired_taint"
      fi
    done

  # watch expired, pause before restarting
  sleep 10
done
```

dystewart:
@larsks I used your logic with the watch and included an initialization loop, which runs once at runtime to ensure all gpu nodes are labeled at runtime. I also tweaked the logic inside the watch to be much more efficient (the taint was firing way to frequently on ALL node updates). The loop now has memory of the prior gpu product label and fires the taint op only if the state has changed

gpu-taint-management/base/taint.sh
```diff
@@ -0,0 +1,37 @@
...
+      else
+        echo "GPU product label removed from $NODE_NAME skipping tainting"
+      fi
+      LAST_PRODUCT[$NODE_NAME]="$PRODUCT"
+    else
+      echo "No change in GPU product label for $NODE_NAME (still '${PRODUCT:-unset}'), skipping"
+    fi
+  done
+  # watch expired, pause before restarting
+  sleep 10
```
jtriley: This looks good overall - running this check every 10s though seems excessive. I would think 5-10 minutes would be better given that this rarely changes unless we're adding nodes or there is some issue with the NVIDIA gpu operator.

jtriley: This could as well be a cronjob that runs once a day. In the case that we're adding new nodes we can just trigger it manually for instant gratification.

jtriley (dismissed):

gpu-taint-management/base/taint.sh
jtriley: Ignore me, I missed the watch option. LGTM. I'll take a look at the deployment on ocp-test cluster today and run some tests.

cluster-scope/bundles/gpu-taint-management/taint.sh
```diff
@@ -0,0 +1,37 @@
...
+    | .metadata.name + " " + .metadata.labels["nvidia.com/gpu.product"]
+  ' |
+while read NODE PRODUCT; do
+    echo "Tainting existing $NODE with $PRODUCT"
+    kubectl taint node "$NODE" "nvidia.com/gpu.product=$PRODUCT:NoSchedule" --overwrite=false
+done
+
+declare -A LAST_PRODUCT
+
+while true; do
```
jtriley, line 20: This loop doesn't appear to handle the case that the gpu.product label exists, hasn't changed, but the taint was somehow removed/lost. During testing I created such a condition when I manually removed the taints and they did not come back until I deleted the deployment pods in order to engage the first part of this script.

jtriley (changes requested):

gpu-taint-management/base/deployment.yaml
```diff
@@ -0,0 +1,38 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: gpu-product-taint
+spec:
+  replicas: 2
```
jtriley: Also, is there a reason we have replicas: 2 here? That seems like it might have the potential to create some race conditions in your script.

dystewart: fixed to 1 replica

cluster-scope/bundles/gpu-taint-management/taint.sh
dystewart: @jtriley Fixed the logic to track the label and taint to cover the case which you mentioned. Now works when labels are added/when taints are removed manually.

cluster-scope/bundles/gpu-taint-management/taint.sh
```diff
@@ -0,0 +1,45 @@
...
+
+# GPU Taint Watcher Script
+
+# Initial tainting of GPU nodes
+kubectl get nodes -l node-role.kubernetes.io/worker -o json |
+jq -r '
+  .items[]
+  | select(.metadata.labels["nvidia.com/gpu.product"] != null)
+  | .metadata.name + " " + .metadata.labels["nvidia.com/gpu.product"]
+' | while read NODE PRODUCT; do
```
larsks, line 11: FWIW, you could avoid using `jq` by doing something like this:

```
kubectl get nodes \
  -l node-role.kubernetes.io/worker \
  -l nvidia.com/gpu.product \
  -o custom-columns='NODE:.metadata.name,PRODUCT:.metadata.labels.nvidia\.com/gpu\.product' --no-headers |
while read NODE PRODUCT; do
```

Not suggesting you need to change it; just offering a suggestion for more effective use of `kubectl`.

cluster-scope/bundles/gpu-taint-management/taint.sh
```diff
@@ -0,0 +1,45 @@
...
+  kubectl taint node "$NODE" "nvidia.com/gpu.product=$PRODUCT:NoSchedule" --overwrite=false 2>/dev/null
+done
+
+# Declare associative array to track last seen GPU product per node
+declare -A LAST_PRODUCT
+
+# Continuous monitoring loop
+while true; do
+  kubectl get nodes -l node-role.kubernetes.io/worker -o json |
+  jq -r '.items[].metadata.name' | while read NODE_NAME; do
```
larsks, line 22: Using `-o name` gets you `kind/name` output, allow you to do this:

```suggestion
  kubectl get nodes -l node-role.kubernetes.io/worker -o name | cut -f2 -d/ |
  while read NODE_NAME; do
```

cluster-scope/bundles/gpu-taint-management/taint.sh
```diff
@@ -0,0 +1,45 @@
...
+    PRODUCT=$(kubectl get node "$NODE_NAME" -o jsonpath="{.metadata.labels['nvidia\.com/gpu\.product']}" 2>/dev/null)
+
+    # Skip nodes without GPU product label
+    if [ -z "$PRODUCT" ]; then
+      echo "No GPU product label on $NODE_NAME, skipping"
+      continue
+    fi
```
larsks, line 29: Instead of listing all workers and then ignoring those without the `nvidia.com/gpu.product` label, why not just add `-l nvidia.com/gpu.product` to the first `kubectl` invocation?

```
kubectl get nodes -l node-role.kubernetes.io/worker -l nvidia.com/gpu.product -o name | cut -f2 -d/ |
while read NODE_NAME; do
    # Check current taints on the node
    CURRENT_TAINT=$(kubectl get node "$NODE_NAME" -o jsonpath="{.spec.taints[?(@.key=='nvidia.com/gpu.product')].value}" 2>/dev/null)
    ...
done
```

cluster-scope/bundles/gpu-taint-management/taint.sh
```diff
@@ -0,0 +1,45 @@
...
+
+    # Check current taints on the node
+    CURRENT_TAINT=$(kubectl get node "$NODE_NAME" -o jsonpath="{.spec.taints[?(@.key=='nvidia.com/gpu.product')].value}" 2>/dev/null)
+
+    # Apply taint if label changed or taint is missing
+    if [[ "${LAST_PRODUCT[$NODE_NAME]}" != "$PRODUCT" || "$CURRENT_TAINT" != "$PRODUCT" ]]; then
+      echo "Applying taint for GPU product '$PRODUCT' on $NODE_NAME (current taint: '${CURRENT_TAINT:-none}')"
+      kubectl taint node "$NODE_NAME" "nvidia.com/gpu.product=$PRODUCT:NoSchedule" --overwrite=false 2>/dev/null
```
larsks, line 37: If the label has changed, presumably the node has an existing taint. Will this command with `--overwrite=false` create a new taint, leaving the old one in place, or will it fail? Both results seem problematic.

joachimweyl:
@jtriley, @dystewart added the suggestions you made, are there any other changes you want made to this PR?

joachimweyl:
@jtriley will re-review soon and make it live after re-testing on ocp-test. Likely next week given that I'm out today and tomorrow

joachimweyl:
@dystewart, re-review done, please make appropriate changes.

PR #756, opened 2025-08-15
Enable RHOAI components in edu cluster

IsaiahStapleton:
This adds the same DataScienceCluster spec that prod cluster uses which enables different RHOAI components. One of which, Kueue, is necessary for running Jonathan's GPU course.

### DIFF
--- a/cluster-scope/overlays/nerc-ocp-edu/rhoai/datasciencecluster.yaml
+++ b/cluster-scope/overlays/nerc-ocp-edu/rhoai/datasciencecluster.yaml
@@ -5,10 +5,6 @@ metadata:
 spec:
   components:
     codeflare:
-      managementState: Removed
-    dashboard:
-      managementState: Managed
-    datasciencepipelines:
       managementState: Managed
     kserve:
       managementState: Managed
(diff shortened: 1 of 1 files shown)


cluster-scope/overlays/nerc-ocp-edu/rhoai/datasciencecluster.yaml
```diff
@@ -21,15 +17,22 @@ spec:
...
+      managementState: Managed
       registriesNamespace: rhoai-model-registries
-    ray:
-      managementState: Removed
     trustyai:
+      managementState: Managed
+    ray:
+      managementState: Managed
+    kueue:
       managementState: Removed
```
DanNiESh: Should it be `Managed`?

IsaiahStapleton: Not sure how I missed that lol. Yes it should be.

IsaiahStapleton: Didn't mean to close the issue, but this has been fixed Kueue is set to managed.

DanNiESh (approved):

tssala23 (approved):

PR #768, opened 2025-08-21
Upgrade Loki and Logging from version 5 to 6

computate:
The upgrade to OpenShift Logging version 6.0 removes the requirement for the Red Hat ElasticSearch operator. We remove ElasticSearch and Jaeger operators. We add the new OpenShift Observability Operator to the obs cluster which enables a new UiPlugin resource definition for the logging dashboard in the console.

Logging UI comes back online with these changes! 
image


schwesig (approved):

larsks (approved):
I've been seeing the deprecation messages for ElasticSearch for a while; glad we're upgrading :)

PR #770, opened 2025-08-25
Upgrade obs cluster from 4.15.32 to 4.16.44 and logging 6.1

computate:
We are upgrading the obs cluster to catch up to the infra, prod, and edu
clusters. At the same time, we need to upgrade Logging and Loki
operators from 6.0 to 6.1.

Relates to https://github.com/nerc-project/operations/issues/1251

### DIFF
--- a/cluster-scope/base/operators.coreos.com/subscriptions/cluster-logging/subscription.yaml
+++ b/cluster-scope/base/operators.coreos.com/subscriptions/cluster-logging/subscription.yaml
@@ -4,7 +4,7 @@ metadata:
   name: cluster-logging
   namespace: openshift-logging
 spec:
-  channel: stable-6.0
+  channel: stable-6.1
   installPlanApproval: Automatic
   name: cluster-logging
   source: redhat-operators
--- a/cluster-scope/base/operators.coreos.com/subscriptions/loki-operator/subscription.yaml
+++ b/cluster-scope/base/operators.coreos.com/subscriptions/loki-operator/subscription.yaml
@@ -4,7 +4,7 @@ metadata:
   name: loki-operator
   namespace: openshift-operators-redhat
 spec:
-  channel: stable-6.0
+  channel: stable-6.1
   installPlanApproval: Automatic
   name: loki-operator
   source: redhat-operators
--- a/cluster-scope/overlays/nerc-ocp-obs/clusterversion.yaml
+++ b/cluster-scope/overlays/nerc-ocp-obs/clusterversion.yaml
@@ -3,7 +3,7 @@ kind: ClusterVersion
 metadata:
   name: version
 spec:
-  channel: stable-4.15
+  channel: stable-4.16
   desiredUpdate:
-    version: 4.15.32
+    version: 4.16.44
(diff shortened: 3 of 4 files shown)


larsks (approved):

cluster-scope/base/operators.coreos.com/subscriptions/cluster-logging/subscription.yaml
```diff
@@ -4,7 +4,7 @@ metadata:
   name: cluster-logging
   namespace: openshift-logging
 spec:
-  channel: stable-6.0
+  channel: stable-6.1
```
larsks, line 7: It looks like the cluster-logging operator is installed on most of our clusters. Changing the version here will upgrade the cluster logging operator *everywhere*. Is that intentional? That already happened in d079066d9e6dbd24e1cba06a46f03156ca2fe71c with the change from 5.8 to 6.0.

jtriley (approved):

cluster-scope/base/operators.coreos.com/subscriptions/cluster-logging/subscription.yaml
computate: Good question @larsks , you'll notice that: 
- [logging operator on nerc-ocp-infra](https://github.com/OCP-on-NERC/nerc-ocp-config/blob/main/cluster-scope/overlays/nerc-ocp-infra/kustomization.yaml#L95-L101) is already at `stable-6.3`. 
- [logging operator on nerc-ocp-prod](https://github.com/OCP-on-NERC/nerc-ocp-config/blob/main/cluster-scope/overlays/nerc-ocp-prod/kustomization.yaml#L145-L151) is already at `stable-6.2`. 
- [logging operator on nerc-ocp-edu](https://github.com/OCP-on-NERC/nerc-ocp-config/blob/main/cluster-scope/overlays/nerc-ocp-edu/kustomization.yaml#L135-L141) is already at `stable-6.3`. 

This is because the versions less than `stable-6.1` are no longer available in the more recent versions of OpenShift. Although since this logging operator is configured in [cluster-scope/overlays/common](https://github.com/OCP-on-NERC/nerc-ocp-config/blob/main/cluster-scope/overlays/common/kustomization.yaml#L18), should this not be part of common, and each cluster should be able to opt in to logging per overlay? 
image

schwesig (approved):

cluster-scope/base/operators.coreos.com/subscriptions/cluster-logging/subscription.yaml
jtriley: It's a good question and brings up a larger issue with having versions defined in base for all operators/subscriptions for all clusters/overlays. We can discuss during one of the weekly meetings but in the interest of moving forward with the production clusters I'm going to merge this for now.

computate:
I added this acknowledgement of kubernetes API removals to advance the cluster upgrade. 

```bash
oc --as system:admin -n openshift-config patch cm admin-acks --patch '{"data":{"ack-4.15-kube-1.29-api-removals-in-4.16":"true"}}' --type=merge
```

See this article: https://access.redhat.com/articles/7031404

PR #780, opened 2025-09-03
Add 6:05am status alerts for nerc-ocp-edu cluster

### LINKED ISSUE #1201: Create the Slack Alert for GPU Class
## Motivation
This is a sub-issue of #1184 

## Completion Criteria
Set up alerts for critical resource thresholds (e.g., GPU memory exhaustion, high error rates) to notify instructors/admins. Gather alerting information from all students namespaces and send them to Slack for professors/TAs to monitor.

## Description
- [ ] First step to resolve the issue

## Completion dates
Desired - 2025-08-20
Required - TBD

schwesig:
fixes https://github.com/nerc-project/operations/issues/1201

Add daily status alerts for nerc-ocp-edu at 6:05am targeting bu-cs599-* namespaces with resource quotas:

- limits.cpu 4
- limits.ephemeral-storage 5368709120 (5 GiB)
- limits.memory 13958643712 (13,312 MiB)
- persistentvolumeclaims 2
- requests.nvidia.com/gpu 1
- requests.storage 21474836480 (20GiB)

quotas will be received by type='hard' and are not hard coded. the numbers here are just for orientation and as set by now.

### DIFF
--- a/acm-observability/overlays/nerc-ocp-infra/configmaps/thanos-ruler-custom-rules/configmap.yaml
+++ b/acm-observability/overlays/nerc-ocp-infra/configmaps/thanos-ruler-custom-rules/configmap.yaml
@@ -193,9 +193,37 @@ data:
             cluster: nerc-ocp-prod
             namespace: rhods-notebooks
 
+        # Recording rules for resource quota percentages - edu cluster (all bu-cs599-* namespaces)
+        - record: ope:resourcequota_usage_ratio_edu
+          expr: >-
+            max_over_time(
+              sum(kube_resourcequota{cluster="nerc-ocp-edu",namespace=~"bu-cs599-.*",type="used"}) by (resource)
+            [10m:5m]) /
+            max_over_time(
+              sum(kube_resourcequota{cluster="nerc-ocp-edu",namespace=~"bu-cs599-.*",type="hard"}) by (resource)
+            [10m:5m])
+          labels:
+            cluster: nerc-ocp-edu
+
+        # Recording rule for container count - edu cluster (all bu-cs599-* namespaces)
+        - record: ope:container_count_edu
+          expr: max_over_time(count(kube_pod_container_info{cluster="nerc-ocp-edu",namespace=~"bu-cs599-.*"})[10m:5m])
+          labels:
+            cluster: nerc-ocp-edu
+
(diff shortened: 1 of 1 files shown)


DanNiESh (approved):

acm-observability/overlays/nerc-ocp-infra/configmaps/thanos-ruler-custom-rules/configmap.yaml
```diff
@@ -193,10 +193,38 @@ data:
...
+          labels:
+            cluster: nerc-ocp-edu
+
         # Recording rule for 6am trigger condition
         - record: ope:is_6am_status_time
           expr: (hour()-4+(minute()/60)==6.0)
 
+        # Recording rule for 6:05am trigger condition for edu cluster
+        - record: ope:is_605am_status_time
+          expr: (hour()-4+(minute()/60)==6.083333)
```
larsks, line 226: I have two questions here:

1. Would this be clearer if the expression where `(hour()-4 == 6) and (minutes() == 5)`?
2. How are we going to handle DST changes? On 11/2, these rules will need `hour()-5` instead of `hour()-4`.

larsks (approved):

computate (approved):

acm-observability/overlays/nerc-ocp-infra/configmaps/thanos-ruler-custom-rules/configmap.yaml
schwesig: see my old PRs:
- there will be a mismatch on DST, and I couldn't find an automated solution. twice a year it needs to be changes, or we accpet the status to be one hour off (I would be very happy if someone has an idea)
- about `(hour()-4 == 6) and (minutes() == 5)`: I did this in my first version (last year), but after some more reading, this version was preferred, so I wnated to try it here.
If there is a preference to the old version, I can totally change that back.

schwesig: added info: DST
that is also the reason why I tried to find a solution (recording rule), that is defined once and can be used multiple times. in the old version, there was as often `-4` as there were alerts. now we just need to change it once.
Most important: the 6am idea was just to be ahead of the first class in the morning.
if this is 5am or 7am would not make big of a difference.

PR #781, opened 2025-09-04
Add necessary clusterroles and kueue configurations for gpu class

DanNiESh:
Add the following changes to edu cluster:
1. The GPU class uses GPUs by submitting batch jobs to the Kueue queue system. Currently, the nominal quota for the v100, a100, and h100 GPUs is set to 4, 0, and 0, respectively. The quota for the a100 and h100 GPUs will be increased later in the class.
2. Add node-reader, pod-reader and clusterqueue-reader clusterroles: students need access to list GPU nodes, get pods from all namespaces that are running on GPUs. They also need to view kueue clusterqueues.


larsks (approved):

tssala23 (approved):

memalhot (approved):

jtriley (approved):

jtriley (changes requested):
This looks good but you've added clusterroles and not clusterrolebindings. Could you please move the clusterroles to a separate folder in the overlay and add the clusterrolebindings with subjects in the clusterrolebindings folder instead?

cluster-scope/overlays/nerc-ocp-edu/clusterrolebindings/gpu-class-clusterroles.yaml
```diff
@@ -0,0 +1,26 @@
...
+  name: kueue-clusterqueue-reader
+rules:
+  - apiGroups: ["kueue.x-k8s.io"]
+    resources: ["clusterqueues"]
+    verbs: ["get", "list", "watch"]
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: node-reader
```
jtriley: We already have a clusterrole called `nerc-node-reader` which could be used instead of adding a new/duplicate one.

larsks:
> This looks good but you've added clusterroles and not clusterrolebindings. Could you please move the clusterroles to a separate folder in the overlay and add the clusterrolebindings with subjects in the clusterrolebindings folder instead?

@jtriley what if we just rename `clusterrolebindings` to `rbac`?

jtriley:
> > This looks good but you've added clusterroles and not clusterrolebindings. Could you please move the clusterroles to a separate folder in the overlay and add the clusterrolebindings with subjects in the clusterrolebindings folder instead?
> 
> @jtriley what if we just rename `clusterrolebindings` to `rbac`?

Sounds good but we still need to add the actual clusterrolebindings.

DanNiESh:
Update: renamed clusterrolebindings to rbac and removed `node-reader` clusterrole

larsks:
> Sounds good but we still need to add the actual clusterrolebindings.

I wasn't sure if we did or not, or if we are referencing these roles in namespaced rolebindings somewhere.

DanNiESh:
> > Sounds good but we still need to add the actual clusterrolebindings.
> 
> I wasn't sure if we did or not, or if we are referencing these roles in namespaced rolebindings somewhere.

These clusterrolebindings are created [here](https://github.com/OCP-on-NERC/BU-RHOAI/blob/main/gpu-class/clusterrolebinding.yaml). Class group users and their notebook service accounts are added to the roles [here](https://github.com/OCP-on-NERC/BU-RHOAI/blob/9f6311f69ca17807010379954dbaa531f9240f7b/gpu-class/gpu-class-setup.sh#L70). Service accounts are added dynamically. I don't think we want to include these crbs in this repo because it will overwrite the newly added service accounts.

jtriley (approved):

jtriley:
@DanNiESh Just FYI I think this will need an update in order to use `nerc-node-neader` clusterrole:
https://github.com/OCP-on-NERC/BU-RHOAI/blob/9f6311f69ca17807010379954dbaa531f9240f7b/gpu-class/gpu-class-setup.sh#L75

PR #791, opened 2025-09-11
add eck operator back to ocp-prod cluster

jtriley:
I mistakenly removed the elasticsearch-eck-operator bundle and pruned the resources after `oc adm upgrade` was complainging about elasticsearch csv during the last upgrade to 4.19. Turns out there were two different elasticsearch operators in use and only the redhat subscription was deprecated/removed.


jtriley:
This is already live on nerc-ocp-prod.

computate (approved):

Milstein (approved):

PR #798, opened 2025-09-24
refactor: Clean up metrics regex in allowlist

schwesig:
- Add ^ anchor for better regex specificity, ensures patterns match from the beginning of metric names
- Remove unnecessary parentheses which created capture groups but added no functionality
- Consolidate kube_node, kube_pod, kube_job patterns using alternation
- Fix acm_managed_cluster pattern (removed redundant wildcard)

The ^ anchor is more efficient, explicitly anchors the pattern to the start of the string rather than creating unnecessary capture groups.


computate (approved):

jimmysway (approved):

larsks (approved):
👍

PR #817, opened 2025-10-24
ClusterRoles with aggregate-to-edit for zookeeper and solr

computate:
For AI Telemetry in development in nerc-ocp-prod and obs cluster, we are migrating away from proprietary bitnami images to the Zookeeper and Solr Operators, which are already available. We just need these aggregate-to-edit role bindings for users and service accounts of these operators to work in their namespace.


IsaiahStapleton (approved):

schwesig (approved):

larsks (approved):
Would there be any benefit in also adding read-only aggregate-to-view roles?

PR #818, opened 2025-10-24
feat(grafana): add image renderer plugin to obs cluster

schwesig:
Enable automated high-resolution report generation for Grafana
in the observability cluster to support RHRQ and operational needs.

Why:
1. For RHRQ (Red Hat Quarterly Reviews), we need high-resolution
   screenshots since the quality from regular screenshots is not
   good enough for presentations and documentation that we have to
   provide to stakeholders
2. In the future we will need regular reports and slides, so it makes
   sense to automate this now instead of doing manual screenshots each
   time till we implement it later

Benefits:
The grafana-image-renderer provides server-side rendering capabilities
for panels and dashboards, so we can generate PNG images through the
Grafana API automatically. This is becomming very useful for automation
workflows where we need consistent, high-quality visuals without manual
work or browser-based rendering each time.

Changes:
1. Added grafana-image-renderer as separate Deployment with dedicated
   container (grafana/grafana-image-renderer:3.11.3) running on port 8081
   with resource limits (CPU: 500m, Memory: 512Mi) to handle Chromium
   browser rendering workloads
2. Created Service for image renderer to expose rendering endpoint
3. Configured Grafana with rendering.server_url, rendering.callback_url,
   and rendering.renderer_token for secure communication
4. Applied OpenShift security context constraints (runAsNonRoot, drop ALL
   capabilities, seccompProfile) to ensure compatibility with restricted SCC
5. Added ExternalSecret to inject renderer token from Vault path:
   nerc/nerc-ocp-obs/grafana/renderer-token
6. Used explicit env variable with secretKeyRef instead of envFrom to avoid
   Grafana Operator reconcile conflicts

Security Improvements:
1. Authentication Token (CVE-2022-31176 mitigation):
   - Configured renderer_token in Grafana (GF_RENDERING_RENDERER_TOKEN)
   - Configured AUTH_TOKEN in image renderer service
   - Token stored in Vault and injected via ExternalSecret operator
   - Prevents unauthorized file disclosure (CVSS 8.3 HIGH vulnerability)
   - Required for renderer v3.6.1+ to avoid security issues
   - Both services must use identical token for authentication to work

2. NetworkPolicy Implementation:
   - Restricts ingress traffic to renderer pod on port 8081/TCP
   - Only Grafana pods (app=grafana label) can access renderer service
   - Follows OpenShift NetworkPolicy best practices (default deny-all)
   - Prevents unauthorized access from other pods in namespace

This approach uses the remote rendering service pattern instead of
plugin mode, which avoids dependency issues with missing libraries
in the Grafana container and provides better security isolation.

Technical Details:
- Image: grafana/grafana-image-renderer:3.11.3
- Port: 8081/TCP (ClusterIP Service)
- Vault Secret Path: nerc/nerc-ocp-obs/grafana/renderer-token
- Required Token Length: 64 characters (32 bytes hex recommended)
- Grafana Version Requirement: v9.1.2+ (or v9.0.8+, v8.5.11+, v8.4.11+, v8.3.11+)
- Renderer Version Requirement: v3.6.1+

Findings:
Grafana Operator v5 Reconcile Behavior:
   During deployment I discovered that the Grafana Operator v5 overwrites
   or doesn't fully respect the grafana-renderer-token when added via envFrom
   secretRef. The operator reconciles the deployment regularly and removed
   the renderer token from the pod configuration, even though it was in the
   Grafana CR spec.

   Solution - Explicit env with secretKeyRef:
       Instead of using envFrom with secretRef for the renderer token, I use
       explicit env variable with valueFrom.secretKeyRef. This pattern is more
       stable with the Grafana Operator because it treats explicit env variables
       different than envFrom during reconciliation. The operator respects
       individual env entries with secretKeyRef better than full secret imports
       via envFrom.

   Configuration pattern:
   ```yaml
   env:
     - name: GF_RENDERING_RENDERER_TOKEN
       valueFrom:
         secretKeyRef:
           name: grafana-renderer-token
           key: GF_RENDERING_RENDERER_TOKEN
   ```

   This is the recommended approach for operator-managed deployments when
   specific environment variables need to persist across reconcile cycles.

Testing:
After deployment, verify rendering with:
  curl -fSL -H "Authorization: Bearer $TOKEN" \
    "$GRAFANA_URL/render/d/<dashboard-uid>?width=3840&height=2160&scale=2"

Goal:
Enable image rendering in Grafana for automated report generation,
supporting both quarterly review materials and scheduled operational
reports with consistent output quality and security.

References:
- Deploying Grafana on OpenShift 4: https://cloud.redhat.com/experts/o11y/ocp-grafana/
- Grafana Image Renderer Plugin: https://grafana.com/grafana/plugins/grafana-image-renderer/
- Grafana Image Renderer GitHub: https://github.com/grafana/grafana-image-renderer
- Remote Rendering Setup: https://grafana.com/docs/grafana/latest/setup-grafana/image-rendering/
- CVE-2022-31176 Security Advisory: https://grafana.com/security/security-advisories/cve-2022-31176/
- OpenShift NetworkPolicy: https://docs.redhat.com/en/documentation/openshift_container_platform/4.12/html/networking/network-policy
- Grafana Operator: https://github.com/grafana/grafana-operator
- Grafana Operator Credential Secret: https://grafana.github.io/grafana-operator/docs/examples/credential_secret/readme/
- Kubernetes NetworkPolicy: https://kubernetes.io/docs/concepts/services-networking/network-policies/

### DIFF
--- a/grafana/overlays/nerc-ocp-obs/grafanas/grafana-image-renderer-service.yaml
+++ b/grafana/overlays/nerc-ocp-obs/grafanas/grafana-image-renderer-service.yaml
@@ -1,0 +1,16 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: grafana-image-renderer
+  namespace: grafana
+  labels:
+    app: grafana-image-renderer
+spec:
+  ports:
+    - port: 8081
+      targetPort: 8081
+      protocol: TCP
+      name: http
+  selector:
+    app: grafana-image-renderer
+  type: ClusterIP
--- a/grafana/overlays/nerc-ocp-obs/grafanas/grafana-image-renderer-service.yaml
+++ b/grafana/overlays/nerc-ocp-obs/grafanas/grafana-image-renderer-service.yaml
@@ -8,7 +8,7 @@ metadata:
 spec:
   ports:
     - port: 8081
-      targetPort: 8081
+      targetPort: http
       protocol: TCP
       name: http
   selector:
--- a/grafana/overlays/nerc-ocp-obs/grafanas/grafana.yaml
+++ b/grafana/overlays/nerc-ocp-obs/grafanas/grafana.yaml
@@ -29,6 +29,10 @@ spec:
       scopes: openid email groups profile
     log:
       mode: console
+    rendering:
+      server_url: 'http://grafana-image-renderer:8081/render'
+      callback_url: 'http://grafana:3000/'
+      renderer_token: '${GF_RENDERING_RENDERER_TOKEN}'
   deployment:
     spec:
       template:
@@ -36,6 +40,12 @@ spec:
           containers:
             - image: 'grafana/grafana:11.3.0'
               name: grafana
+              env:
+                - name: GF_RENDERING_RENDERER_TOKEN
+                  valueFrom:
+                    secretKeyRef:
+                      name: grafana-renderer-token
+                      key: GF_RENDERING_RENDERER_TOKEN
               envFrom:
                 - secretRef:
                     name: oauth-client-secret
--- a/grafana/overlays/nerc-ocp-obs/externalsecrets/grafana-renderer-token.yaml
+++ b/grafana/overlays/nerc-ocp-obs/externalsecrets/grafana-renderer-token.yaml
@@ -1,0 +1,20 @@
+apiVersion: external-secrets.io/v1beta1
+kind: ExternalSecret
+metadata:
+  name: grafana-renderer-token
+  namespace: grafana
+spec:
+  secretStoreRef:
+    name: nerc-cluster-secrets
+    kind: ClusterSecretStore
+  target:
+    name: grafana-renderer-token
+  data:
+  - secretKey: GF_RENDERING_RENDERER_TOKEN
+    remoteRef:
+      key: nerc/nerc-ocp-obs/grafana/renderer-token
+      property: TOKEN
+  - secretKey: AUTH_TOKEN
+    remoteRef:
+      key: nerc/nerc-ocp-obs/grafana/renderer-token
+      property: TOKEN
--- a/grafana/overlays/nerc-ocp-obs/externalsecrets/kustomization.yaml
+++ b/grafana/overlays/nerc-ocp-obs/externalsecrets/kustomization.yaml
@@ -2,3 +2,4 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 resources:
 - oauth-client-secret.yaml
+- grafana-renderer-token.yaml
--- a/grafana/overlays/nerc-ocp-obs/grafanas/grafana-image-renderer-deployment.yaml
+++ b/grafana/overlays/nerc-ocp-obs/grafanas/grafana-image-renderer-deployment.yaml
@@ -1,0 +1,65 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: grafana-image-renderer
+  namespace: grafana
+  labels:
+    app: grafana-image-renderer
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: grafana-image-renderer
+  template:
+    metadata:
+      labels:
+        app: grafana-image-renderer
+    spec:
+      securityContext:
+        runAsNonRoot: true
+        seccompProfile:
+          type: RuntimeDefault
+      containers:
+        - name: grafana-image-renderer
+          image: grafana/grafana-image-renderer:3.11.3
+          ports:
+            - containerPort: 8081
+              name: http
+              protocol: TCP
+          env:
+            - name: BROWSER_TZ
+              value: UTC
+            - name: ENABLE_METRICS
+              value: 'true'
+            - name: HTTP_PORT
+              value: '8081'
+          envFrom:
+            - secretRef:
+                name: grafana-renderer-token
+          livenessProbe:
+            httpGet:
+              path: /
+              port: http
+            initialDelaySeconds: 30
+            periodSeconds: 10
+            timeoutSeconds: 5
+            failureThreshold: 3
+          readinessProbe:
+            httpGet:
+              path: /
+              port: http
+            initialDelaySeconds: 10
+            periodSeconds: 5
+            timeoutSeconds: 3
+          resources:
+            limits:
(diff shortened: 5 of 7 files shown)


grafana/overlays/nerc-ocp-obs/grafanas/grafana.yaml
```diff
@@ -36,6 +36,9 @@ spec:
           containers:
             - image: 'grafana/grafana:11.3.0'
               name: grafana
+              env:
```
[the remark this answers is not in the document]
schwesig: The resource limits are already there actually, in the separate grafana-image-renderer Deployment (lines 51-57 in
grafana-image-renderer-deployment.yaml) — with exactly the values.
At first I used the plugin way, but then I changed to the remote rendering service thing, which already has the proper resource limits included.

schwesig:
| Komponente     | Our Version | Min. Fix Version | Status   |
  |----------------|----------------|------------------|----------|
  | Grafana Core   | 11.3.0         | v9.1.2+          | check |
  | Image Renderer | 3.11.3         | v3.6.1+          | scheck |

schwesig:
oh boy, this was planned as a quick "activate render plugin", and now became a full stack :-(
sorry

grafana/overlays/nerc-ocp-obs/grafanas/grafana-image-renderer-service.yaml
```diff
@@ -0,0 +1,16 @@
...
+kind: Service
+metadata:
+  name: grafana-image-renderer
+  namespace: grafana
+  labels:
+    app: grafana-image-renderer
+spec:
+  ports:
+    - port: 8081
+      targetPort: 8081
```
larsks: You make your life easier if you use the port name instead of number for the targetPort (because that permits you to change ports in the pod without needing to update the service).

Mostly just an FYI; doesn't need changing.

larsks (approved):

grafana/overlays/nerc-ocp-obs/grafanas/grafana-image-renderer-service.yaml
schwesig: argh, yes, you are so right. thanks for that feedback.

schwesig: updated

naved001 (approved):

schwesig:
these two fixes/features
- https://github.com/OCP-on-NERC/nerc-ocp-config/pull/820
- https://github.com/OCP-on-NERC/nerc-ocp-config/pull/821

schwesig:
Closing with kind of a readme

# Grafana Renderer now deployed on OBS

We've deployed the Grafana image rendering service on the observability cluster.  
You can now generate high-resolution PNG exports of dashboards and panels via API/curl.

**PR:** https://github.com/OCP-on-NERC/nerc-ocp-config/pull/818

## Usage example

```bash
curl -fSL \
  -H "Authorization: Bearer $GRAFANA_TOKEN" \
  "https://grafana.apps.obs.nerc.mghpcc.org/render/d-solo/DASHBOARD_UID/dashboard-name?from=now-7d&to=now&panelId=14&width=3840&height=2160&scale=2" \
  -o output.png
```

```bash
curl -fSL \
  -H "Authorization: Bearer $GRAFANA_TOKEN" \
  "https://grafana.apps.obs.nerc.mghpcc.org/render/d-solo/bewz2he55l728b_v2/ope-edu-v2-jonathan?from=now7d&to=now&panelId=1&width=1920&height=1080&scale=2" \
  -o "panel-1-bewz2he55l728b_v2_scale2.png"
```

## Getting your `GRAFANA_TOKEN`

Grafana UI → **Users and access** → **Service accounts** → **Add Service Account** → **Add Service Account token**

## Single Panel

- `/render/d-solo/{uid}/{slug}?panelId={id}&width={w}&height={h}&scale={s}`
- `/render/d-solo/...` with `panelId` parameter — renders a single panel

## Full Dashboard

- `/render/d/{uid}/{slug}?width={w}&height={h}&scale={s}&kiosk`
- `/render/d/...` with `kiosk` parameter — renders entire dashboard without sidebar/menus
- `height=-1` (workaround) may enable full-page rendering with dynamic dashboard height, but it is undocumented and inconsistent across versions. [1](https://github.com/grafana/grafana-image-renderer/issues/538#issue-2377830135) [2](https://github.com/scylladb/scylla-monitoring/issues/2324)

## Notes

- **Scale parameter** doubles the rendering resolution  
  → e.g., `scale=2` at 3840×2160 renders at **7680×4320**.
- Higher values may timeout on complex dashboards  
  → up to ~25s should be fine, timeout can be increased manually.
- If fonts/titles/legends look too small:  
  → render at a smaller base size (e.g., 1920×1080) and increase `scale`.
- For long dashboards, use **very high height values** (e.g., 8000px+).

schwesig:
For full-size dashboards, set:

`height = -1`

This captures a snapshot of the entire dashboard.

schwesig:
in url
&timeout=120

on haproxy.router
`oc annotate route grafana -n grafana haproxy.router.openshift.io/timeout=600s --overwrite --as=system:admin`

schwesig:
just wanted to share a finding that might be useful if someone else runs into render timeouts.                                                              
																		  
TL;DR: the dashboard auto-refresh interval was the actual problem.            
																			
@IsaiahStapleton  was trying to render a dashboard to a high-res PNG via the Grafana /render/ API and always got timeouts (504, 500, or just 0 bytes after 5-10 minutes).

i thought it was a memory or route timeout issue – and yes, the renderer pod was also OOMkilled (512Mi limit, increased to 2Gi) and the HAProxy route had only 30s timeout (increased to 600s).

but even after all this, the render was still hanging.

the real fix: i changed the dashboard auto-refresh from 15s to 15min and saved the dashboard.
after that, renders finish in under 20 seconds, at any resolution.                                                                   
																		  
why: the headless browser the renderer uses loads the dashboard and waits for it to "settle" before taking the  screenshot. with a 15s refresh interval, the browser keeps firing new Prometheus query waves continuously.
it never settles, queries get aborted, and the render either times out or crashes.   

so if you want to render a dashboard: set the refresh interval to something high (15min or off), save it, render, done.

PR #843, opened 2026-01-14
fix(obs): correct nerc-logo namespace

schwesig:
The Console configuration expects ConfigMaps to exist in openshift-config namespace, not openshift-console. The branding was not working because the nerc-logo ConfigMap was placed in wrong namespace. This was discovered when testing the Console customization and logo was not appearing correctly.

- Changed namespace from openshift-console to openshift-config in nerc-logo.yaml

- This allows the Console to correctly reference the logo ConfigMap

- Fixes branding issue on OBS cluster

Fixes: #842, #841

Triggered by: #842, #841

Connected to: n/a


computate (approved):

IsaiahStapleton (approved):

schwesig:
**Closing this PR without merge.**

The fix here is correct about the namespace problem (`openshift-console` → `openshift-config`), but it creates a Kustomize error:

`may not add resource with an already registered id: ConfigMap.v1.[noGrp]/nerc-logo.openshift-config`

Problem is: the base logo ConfigMap (in `cluster-scope/base/core/configmaps/nerc-logo/`) already comes in through `bundles/console` → `common`.

When we add another ConfigMap with same name and namespace as resource, Kustomize sees it as duplicate.

Better solution is to use a **strategic merge patch** instead of separate resource. This is done in #844.

PR #850, opened 2026-01-30
ocp-test: add udev rules for H100 nodes

jtriley:
This adds udev rules to configure the 6x Mellanox NICs on the Lenovo SD665NV3 H100 nodes. These nodes live in ESI and only 5x of the NICs are cabled/connected: 1x connectx-6 card and 4x connectx-7 cards. We configure the 1x disconnected connectx-6 card given that we need to configure a bond similar to the rest of the cluster. The bond network setup is functional but only has one active NIC.


tssala23 (approved):

larsks (approved):

larsks (changes requested):
~~Spotted what looks like a typo.~~

Never mind.

larsks (approved):

aabaris (approved):

jtriley:
> ~Spotted what looks like a typo.~
> 
> Never mind.

Well, whatever it was I did rearrange the devices and forgot to fix the naming. Just force pushed a minor change for that 😄

PR #853, opened 2026-02-03
add nvidia-network-operator bundle and install on nerc-ocp-test

jtriley:
This patch does the following:

1. Creates a new bundle, `nvidia-network-operator`, for installing the NVIDIA network operator.
2. Adds the nvidia-network-operator bundle to the `nerc-ocp-test` cluster as part of testing RDMA.
3. Adds initial NIC policy for `nerc-ocp-test` RDMA testing.

### DIFF
--- a/cluster-scope/base/core/namespaces/nvidia-network-operator/kustomization.yaml
+++ b/cluster-scope/base/core/namespaces/nvidia-network-operator/kustomization.yaml
@@ -1,0 +1,4 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+- namespace.yaml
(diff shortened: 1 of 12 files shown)


tssala23 (approved):
lgtm, then there would be a top level dir fo the nic cluster policy

jtriley:
> lgtm, then there would be a top level dir fo the nic cluster policy

Yeah I was debating on adding that as a separate effort but will include it here before merging.

jtriley:
> lgtm, then there would be a top level dir fo the nic cluster policy

That piece is done now although we might need to adjust the image version for 4.19 vs 4.20 as you pointed out. Looking into that now.

tssala23 (approved):

jtriley:
> > lgtm, then there would be a top level dir fo the nic cluster policy
> 
> That piece is done now although we might need to adjust the image version for 4.19 vs 4.20 as you pointed out. Looking into that now.

Should be fine given that 4.19 also runs RHEL 9.6 (Plow). We can adjust later if we run into issues.

aabaris (approved):

PR #864, opened 2026-02-11
authorino: use stable channel

jtriley:
The ocp-edu cluster RHOAI model-registry-operator-controller-manager pod is having issues fetching a newer version of the authorino API causing it to go into a crashloopbackoff state:
```
failed to get API group resources: unable to retrieve the complete list of server APIs: authorino.kuadrant.io/v1beta3: the server could not find the requested resource
```
This patch removes the need to patch the authorino subscription in the nerc-ocp-prod overlay.

### DIFF
--- a/cluster-scope/base/operators.coreos.com/subscriptions/authorino-operator/subscription.yaml
+++ b/cluster-scope/base/operators.coreos.com/subscriptions/authorino-operator/subscription.yaml
@@ -4,7 +4,7 @@ metadata:
   name: authorino-operator
   namespace: openshift-operators
 spec:
-  channel: tech-preview-v1
+  channel: stable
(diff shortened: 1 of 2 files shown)


tssala23 (approved):

IsaiahStapleton (approved):

jtriley:
Just noting there's a chance this was also the underlying issue with the weird ingress behavior we encountered on the ocp-edu cluster. This latest stable version might no longer use iptables rules within the pod which was the root issue that caused iptables pod to interfere with ingress router pods health checks. I'm monitoring iptables pod logs to see if it mentions authorino after the upgrade.

See: https://github.com/nerc-project/operations/issues/1314#issuecomment-3855933203

PR #865, opened 2026-02-12
Upgrade ocp-test cluster to 4.20

dystewart:
Upgrade ocp-test cluster to 4.20
ocp-test: Provide admin acks for test cluster upgrade
Confirmed removed apis are not being served in cluster
See https://access.redhat.com/articles/7130599

### DIFF
--- a/cluster-scope/overlays/nerc-ocp-test/kustomization.yaml
+++ b/cluster-scope/overlays/nerc-ocp-test/kustomization.yaml
@@ -60,6 +60,7 @@ configMapGenerator:
   behavior: merge
   literals:
     - ack-4.18-kube-1.32-api-removals-in-4.19=true
+    - ack-4.19-admissionregistration-v1beta1-api-removals-in-4.20=true
 - files:
   - config.yaml=configmaps/cluster-monitoring-config.yaml
   name: cluster-monitoring-config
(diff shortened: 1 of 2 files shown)


tssala23 (approved):

cluster-scope/overlays/nerc-ocp-test/kustomization.yaml
```diff
@@ -59,7 +59,7 @@ configMapGenerator:
   namespace: openshift-config
   behavior: merge
   literals:
-    - ack-4.18-kube-1.32-api-removals-in-4.19=true
+    - ack-4.19-admissionregistration-v1beta1-api-removals-in-4.20=true
```
naved001, line 63: when we have upgraded other clusters, we've simply added to this list instead of modifying the previous literal. I don't think it'll matter and I don't know what the right or wrong thing is in this case, but I wanted to point it out.

See example: https://github.com/OCP-on-NERC/nerc-ocp-config/blob/main/cluster-scope/overlays/nerc-ocp-prod/kustomization.yaml#L74-L78

jtriley: Yeah I would leave it in if nothing else for historical purposes.

dystewart: Replaced the previous literal

naved001 (approved):

jtriley (approved):

PR #884, opened 2026-03-11
Improve vLLM Dashboard

IsaiahStapleton:
View dashboard here: https://grafana.apps.obs.nerc.mghpcc.org/d/cffpy3z56d43kd/vllm-2-0

Summary of changes:
- Change Queue Time query from raw rate to average per request. Also add percentiles to this panel.
- Change Max Generation Tokens panel from raw rate to average per request
- Add $window variable for configurable rate intervals
- Replace hardcoded $__rate_interval with $window var on all panels
- Add row sections: Latency, Throughput & Requests, Scheduler & Cache, Token Distribution
- Enable shared crosshair across all panels (graphTooltip: 1)
- Add fill opacity to panels to help differentiate between different lines
- Add table legends with mean/last calcs
- Add multi series tooltips for panels with percentiles
- Add units and axis labels to panels missing them
- Remove hardcoded values: dashboard id, uid, datasource, and model values

### DIFF
--- a/grafana/overlays/nerc-ocp-obs/grafanadashboards/vllm-dashboard.yaml
+++ b/grafana/overlays/nerc-ocp-obs/grafanadashboards/vllm-dashboard.yaml
@@ -37,10 +37,23 @@ spec:
       "description": "Monitoring vLLM Inference Server",
       "editable": true,
       "fiscalYearStartMonth": 0,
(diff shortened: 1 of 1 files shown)


computate (approved):

schwesig (approved):
sorry for being so late in review, my bad, I missed that

PR #894, opened 2026-03-24
disable dashboard

memalhot:
Disable dashboard on production cluster so students no longer have access to the dashboard. Was turned on briefly for testing prod.


larsks (approved):
Approved, but add some sort of description to the commit message (why are we disabling the dashboard?)

DanNiESh (approved):

PR #930, opened 2026-05-22
fix: NAIRR dashboard - replace image!='' filter with container!='' in CPU queries

schwesig:
## Problem

`docker-build` containers in cAdvisor have no `image` label set, so the filter `image!=""` silently drops all their CPU metrics. This causes the **CPU used** and **Estimated power** panels to show "No data" for namespaces running build workloads.

## Fix

Replace `image!=""` with `container!=""` in both CPU queries.

- `container!="POD"` already excludes the pause/sandbox container
- `container!=""` additionally excludes the bare node-level aggregates
- No need for `image!=""` — it only accidentally filters real workloads

Affected panels:
- CPU used (cores)
- Estimated power usage (heuristic)

Relates to nerc-project/operations#1394

### DIFF
--- a/grafana/overlays/nerc-ocp-obs/grafanadashboards/nairr-pilot-project-dashboard.yaml
+++ b/grafana/overlays/nerc-ocp-obs/grafanadashboards/nairr-pilot-project-dashboard.yaml
@@ -76,7 +76,7 @@ spec:
                 },
                 "targets": [
                     {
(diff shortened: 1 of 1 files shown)


schwesig:
before
image

larsks (approved):

IsaiahStapleton (approved):

PR #933, opened 2026-06-02
Add AdminNetworkPolicy for suhruth-test namespace

IsaiahStapleton:
The purpose of this policy is to allow suhruth to work with openclaw agents in the test cluster in a safe way. As he does his work we will likely need to modify this as needed for whatever his setup demands, for example access to specific endpoints. 

I am almost certain this is too restrictive. For one, as is the agent can't call out to the anthropic server to prompt claude.  Also he will likely be using the claw installer developed by Sally: https://github.com/sallyom/claw-installer. Which might require certain other permissions.

### DIFF
--- a/cluster-scope/overlays/nerc-ocp-test/adminnetworkpolicies/restrict-suhruth-test-egress.yaml
+++ b/cluster-scope/overlays/nerc-ocp-test/adminnetworkpolicies/restrict-suhruth-test-egress.yaml
@@ -5,21 +5,21 @@ metadata:
 spec:
   priority: 50
   subject:
-    namespaces:
-      matchLabels:
-        kubernetes.io/metadata.name: suhruth-test
-    podSelector:
-      matchLabels:
-        app: openclaw
+    pods:
+      namespaceSelector:
+        matchLabels:
+          kubernetes.io/metadata.name: suhruth-test
+      podSelector:
+        matchLabels:
+          app: openclaw
   egress:
   # Rule 1: allow DNS lookups to OpenShift DNS so pods can resolve hostnames.
   - name: "allow-dns"
     action: "Allow"
     to:
     - namespaces:
-        namespaceSelector:
-          matchLabels:
-            kubernetes.io/metadata.name: openshift-dns
+        matchLabels:
+          kubernetes.io/metadata.name: openshift-dns
     ports:
     - portNumber:
         protocol: UDP
@@ -31,7 +31,7 @@ spec:
   - name: "allow-kube-api"
     action: "Allow"
     to:
-    - namespaces:
+    - pods:
         namespaceSelector:
           matchLabels:
             kubernetes.io/metadata.name: default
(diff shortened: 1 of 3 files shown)


schwesig (approved):

svuppala2006 (approved):

cluster-scope/overlays/nerc-ocp-test/adminnetworkpolicies/restrict-suhruth-test-egress.yaml
```diff
@@ -0,0 +1,51 @@
...
+  - name: "allow-kube-api"
+    action: "Allow"
+    to:
+    - namespaces:
+        namespaceSelector:
+          matchLabels:
+            kubernetes.io/metadata.name: default
+        podSelector:
+          matchLabels:
+            component: apiserver
```
larsks: I think there may be some errors here:

```
Error from server (BadRequest): error when creating
"restrict-suhruth-test-egress.yaml": AdminNetworkPolicy in version "v1alpha1"
cannot be handled as a AdminNetworkPolicy: strict decoding error: unknown field
"spec.egress[0].to[0].namespaces.namespaceSelector", unknown field
"spec.egress[1].to[0].namespaces.namespaceSelector", unknown field
"spec.egress[1].to[0].namespaces.podSelector", unknown field
"spec.subject.podSelector"
```

Looking at the output of `kubtectl explain adminnetworkpolicy.spec.egress.to.namespaces`, it looks like the only allowable fields here are `matchExpressions` and `matchLabels`:

```
GROUP:      policy.networking.k8s.io
KIND:       AdminNetworkPolicy
VERSION:    v1alpha1

FIELD: namespaces <Object>


DESCRIPTION:
    Namespaces defines a way to select all pods within a set of Namespaces.
    Note that host-networked pods are not included in this type of peer.


    Support: Core

FIELDS:
  matchExpressions	<[]Object>
    matchExpressions is a list of label selector requirements. The requirements
    are ANDed.

  matchLabels	<map[string]string>
    matchLabels is a map of {key,value} pairs. A single {key,value} in the
    matchLabels
    map is equivalent to an element of matchExpressions, whose key field is
    "key", the
    operator is "In", and the values array contains only "value". The
    requirements are ANDed.
```

IsaiahStapleton: @larsks thanks for catching that. When i oc apply the policy that error isn't present but it is present when doing kubectl. I've pushed a new commit to address these issues

larsks:
> Also he will likely be using the claw installer developed by Sally: [sallyom/claw-installer](https://github.com/sallyom/claw-installer?rgh-link-date=2026-06-02T21%3A07%3A50Z). Which might require certain other permissions.

He should be building a custom image outside of openshift and then deploying the image. This both provides a record of how to build the test environment, and it also avoids conflicts with the network policy.

larsks (approved):
This looks like a good start!

PR #946, opened 2026-06-29
Add second Google OAuth IP range for OpenClaw (173.194.0.0/16)

svuppala2006:
## Problem

After PR #945 was merged, OpenClaw still cannot authenticate with Google Vertex AI. Testing revealed that `oauth2.googleapis.com` resolves to **different IPs** depending on where the DNS query originates.

## Root Cause Discovery

DNS resolution varies by location:

**External (my machine):**
```bash
$ host oauth2.googleapis.com
oauth2.googleapis.com has address 142.251.127.95
```

**Internal (from OpenClaw pod):**
```bash
$ oc exec deployment/openclaw -c gateway -- node -e "require('dns').resolve4('oauth2.googleapis.com', (err, addrs) => console.log(err || addrs))"
[ '173.194.45.95' ]
```

PR #945 only added `142.250.0.0/15`, but the pod actually needs `173.194.0.0/16`.

## Solution

Add the second IP range to Rule 6:

```yaml
# Rule 6: Allow Google APIs (Vertex AI + OAuth2 for litellm)
- action: Allow
  name: allow-google-apis
  ports:
    - portNumber:
        port: 443
        protocol: TCP
  to:
    - networks:
        - 216.239.32.0/22    # us-east5-aiplatform.googleapis.com (Vertex AI)
        - 142.250.0.0/15     # oauth2.googleapis.com (Google OAuth range 1)
        - 173.194.0.0/16     # oauth2.googleapis.com (Google OAuth range 2 - resolves from pod)
```

## Verification

```bash
$ whois 173.194.45.95 | grep CIDR
CIDR: 173.194.0.0/16
```

## Impact

- ✅ Allows OpenClaw to authenticate with Google Vertex AI
- ✅ Unblocks all Claude API requests through OpenClaw  
- ✅ No security impact - still restricted to HTTPS on port 443

## Testing

After merge and ArgoCD sync:

```bash
# Test from inside pod
oc exec deployment/openclaw -c gateway -n suhruth-test -- \
  curl -I --connect-timeout 5 https://oauth2.googleapis.com

# Should succeed instead of timing out
```

## Related PRs

- PR #943 - Initial Presidio network policy (merged)
- PR #944 - Fixed Presidio port to 3000 (merged)
- PR #945 - Added first Google OAuth range 142.250.0.0/15 (merged)
- This PR - Adds second Google OAuth range 173.194.0.0/16

---

**Priority:** High - OpenClaw is non-functional  
**Risk:** Low - Only adding one more IP range to existing allow rule

### DIFF
--- a/cluster-scope/overlays/nerc-ocp-test/adminnetworkpolicies/restrict-suhruth-test-egress.yaml
+++ b/cluster-scope/overlays/nerc-ocp-test/adminnetworkpolicies/restrict-suhruth-test-egress.yaml
@@ -105,7 +105,8 @@ spec:
       to:
         - networks:
             - 216.239.32.0/22    # us-east5-aiplatform.googleapis.com (Vertex AI)
-            - 142.250.0.0/15     # oauth2.googleapis.com (Google OAuth - CORRECT IP RANGE)
+            - 142.250.0.0/15     # oauth2.googleapis.com (Google OAuth range 1)
+            - 173.194.0.0/16     # oauth2.googleapis.com (Google OAuth range 2 - resolves from pod)
 
     # Rule 7: Deny everything else
     - action: Deny


cluster-scope/overlays/nerc-ocp-test/adminnetworkpolicies/restrict-suhruth-test-egress.yaml
```diff
@@ -105,7 +105,8 @@ spec:
       to:
         - networks:
             - 216.239.32.0/22    # us-east5-aiplatform.googleapis.com (Vertex AI)
-            - 142.250.0.0/15     # oauth2.googleapis.com (Google OAuth - CORRECT IP RANGE)
+            - 142.250.0.0/15     # oauth2.googleapis.com (Google OAuth range 1)
+            - 173.194.0.0/16     # oauth2.googleapis.com (Google OAuth range 2 - resolves from pod)
```
naved001, line 109: Sorry for barging in and I don't have much context, but why are we allowing giant ranges like /15 and /16? That's like several 100,000 IPs being allowed. This doesn't seem very restrictive to me to be honest.

naved001 (changes requested):
I am hitting the request changes button because I want more clarity before this is merged.

cluster-scope/overlays/nerc-ocp-test/adminnetworkpolicies/restrict-suhruth-test-egress.yaml
svuppala2006: @naved001 No, you are definitely correct in that using big ranges is not very restrictive/secure. The problem is that ANP doesn't have the capabilities to make use of domain names (for networking, it strictly uses IPs). I did find that there is an alternative, which is EgressFirewall. It's similar, but it allows for domain names, is namespace-scoped, and is evaluated after all the other network policies. But, because its evaluated later, some of the policies in the ANP would need to be moved over such as denying all other egress. So, this is definitely an option, but these IP ranges you see are also Google-owned ranges, so they are trustyworthy. However, I understand this isn't best practice, so let me know what you think or if you have any suggestions.

naved001: > I did find that there is an alternative, which is EgressFirewall. It's similar, but it allows for domain names, is namespace-scoped, and is evaluated after all the other network policies. But, because its evaluated later, some of the policies in the ANP would need to be moved over such as denying all other egress. So, this is definitely an option, but these IP ranges you see are also Google-owned ranges, so they are trustyworthy. 

If there's a workaround using a combination of EgressFirewall and ANP then that would be the preferred option. Same goes for reaching out to github.

svuppala2006:
I'm going to work on an alternate solution that won't include these large IP address ranges. Likely, it will be via EgressFirewall, as EgressFirewall allows for domain names and ANP doesn't. I will address this alternate solution in a different PR.

PR #947, opened 2026-07-01
Add EgressFirewall with DNS-based domain filtering for OpenClaw

svuppala2006:
## Summary

Replace broad Google IP ranges (/15, /16) with DNS-based domain filtering using EgressFirewall for better security and maintainability.

**Supersedes:** #946

### 1. Updated AdminNetworkPolicy
- **Before:** Allowed specific Google IP ranges (142.250.0.0/15, 173.194.0.0/16)
- **After:** Allow HTTPS (port 443) to 0.0.0.0/0
- **Why:** EgressFirewall now handles domain-based filtering

### 2. Added EgressFirewall (NEW)
DNS-based domain allowlist:
- `oauth2.googleapis.com`
- `*.aiplatform.googleapis.com`
- `accounts.google.com`
- `github.com`, `*.github.com`

## Security Improvement

| Metric | Before (IP Ranges) | After (EgressFirewall) |
|--------|-------------------|------------------------|
| IPs Allowed | ~200,000 | Only IPs from allowed domains |
| Attack Surface | Entire Google infrastructure | Only 3 Google domains |
| Maintainability | Manual IP updates | Auto-follows DNS |

## Defense in Depth

**Layer 1 - AdminNetworkPolicy (Pod-Level):**
- Only `app=openclaw` pods can make HTTPS requests

**Layer 2 - EgressFirewall (Namespace-Level):**
- Only specific domains allowed
- All other domains denied by default

## Files Changed

- `cluster-scope/overlays/nerc-ocp-test/adminnetworkpolicies/restrict-suhruth-test-egress.yaml`
- `cluster-scope/overlays/nerc-ocp-test/egressfirewalls/suhruth-test-egress.yaml` (NEW)
- `cluster-scope/overlays/nerc-ocp-test/kustomization.yaml`

---

**Related:** nerc-project/operations#1564

### DIFF
--- a/cluster-scope/overlays/nerc-ocp-test/adminnetworkpolicies/restrict-suhruth-test-egress.yaml
+++ b/cluster-scope/overlays/nerc-ocp-test/adminnetworkpolicies/restrict-suhruth-test-egress.yaml
@@ -25,6 +25,12 @@ spec:
         - portNumber:
             port: 5353
             protocol: TCP
+        - portNumber:
+            port: 53
+            protocol: UDP
+        - portNumber:
+            port: 53
+            protocol: TCP
       to:
         - namespaces:
             matchLabels:
@@ -72,30 +78,7 @@ spec:
             matchLabels:
               kubernetes.io/metadata.name: suhruth-test
 
-    # Rule 4: Allow OpenShift OAuth router (for GitHub login flow)
-    - action: Allow
-      name: allow-openshift-oauth
-      ports:
-        - portNumber:
-            port: 443
-            protocol: TCP
-      to:
-        - networks:
-            - 199.94.63.12/32  # *.apps.ocp-test.nerc.mghpcc.org (external router IP)
-            - 10.30.8.22/32    # internal cluster DNS resolution for oauth-openshift.apps
-
-    # Rule 5: Allow GitHub (for OAuth identity provider)
-    - action: Allow
-      name: allow-github
-      ports:
-        - portNumber:
-            port: 443
-            protocol: TCP
-      to:
-        - networks:
-            - 140.82.112.0/20  # GitHub
-
-    # Rule 6: Allow HTTPS egress
+    # Rule 4: Allow HTTPS egress
     # EgressFirewall (namespace-level) restricts which domains can be reached
     # AdminNetworkPolicy (pod-level) restricts which pods can make HTTPS requests
     - action: Allow
@@ -108,7 +91,7 @@ spec:
         - networks:
             - 0.0.0.0/0
 
-    # Rule 7: Deny everything else
+    # Rule 5: Deny everything else
     - action: Deny
       name: deny-all-other-egress
       to:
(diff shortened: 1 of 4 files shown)


cluster-scope/overlays/nerc-ocp-test/adminnetworkpolicies/restrict-suhruth-test-egress.yaml
```diff
@@ -95,17 +95,18 @@ spec:
...
+      name: allow-https-egress
       ports:
         - portNumber:
             port: 443
             protocol: TCP
       to:
         - networks:
-            - 216.239.32.0/22    # us-east5-aiplatform.googleapis.com (Vertex AI)
-            - 142.250.0.0/15     # oauth2.googleapis.com (Google OAuth - CORRECT IP RANGE)
+            - 0.0.0.0/0
```
naved001, line 92: With this rule in place, you don't need specific exceptions in Rule 4 and 5.

cluster-scope/overlays/nerc-ocp-test/egressfirewalls/suhruth-test-egress.yaml
```diff
@@ -0,0 +1,60 @@
...
+  egress:
+  # Rule 1: Allow DNS to OpenShift DNS service
+  - type: Allow
+    to:
+      cidrSelector: 172.30.0.10/32
+    ports:
+      - protocol: UDP
+        port: 53
+      - protocol: TCP
+        port: 53
```
naved001, line 16: The ANP allows port 5353 for DNS, and this one allows 53. So, wouldn't the ANP block connection to port 53 and will never hit this allow rule. Similarly, the ANP would permit access to 5353 but the egressfirewall will block that? I am not sure so just wanted to get some clarity.

svuppala2006 (commented):
Great catches! I've addressed both:

**Comment 1 (redundant rules):** ✅ Removed Rules 4 & 5 (OpenShift OAuth, GitHub) since they're now covered by the broader allow-https-egress rule.

**Comment 2 (DNS port mismatch):** ✅ Added port 53 (TCP/UDP) to the ANP DNS rule. Now both ANP and EgressFirewall allow port 53, ensuring DNS works correctly.

The updated architecture:
- **ANP Rule 1:** Allow DNS on ports 53 + 5353 (both TCP/UDP)
- **EgressFirewall Rule 1:** Allow DNS on port 53 to 172.30.0.10/32

Thanks for the thorough review!

naved001 (approved):

IsaiahStapleton (approved):

IsaiahStapleton:
one nitpicky comment is what Lars mentioned, and its about the comments. Rather than having like rule1...ruleX, you can just remove the numbering and state what the rule is doing so you dont have to fix the numbers every time.

svuppala2006:
> one nitpicky comment is what Lars mentioned, and its about the comments. Rather than having like rule1...ruleX, you can just remove the numbering and state what the rule is doing so you dont have to fix the numbers every time.

Got it. I'll change that on the next PR.

PR #953, opened 2026-07-06
feat: add 5m DCGM recording rules to bridge ACM 300s collection gap

schwesig:
ACM Observability forwards only one point-in-time sample per metric every 300s (observabilityAddonSpec.interval), so short GPU bursts on gauges like DCGM_FI_DEV_GPU_UTIL can be shifted, shortened, or missed entirely in the central Grafana (see the ~8min training run on nerc-ocp-test that showed up as 5min, offset by ~5min).

Counters are unaffected (deltas between snapshots stay correct); gauges lose everything between snapshots. This change pre-aggregates the most relevant DCGM gauges locally on nerc-ocp-test via recording rules:

- nerc:dcgm_gpu_util:{max5m,avg5m}
- nerc:dcgm_gr_engine_active:{max5m,avg5m}
- nerc:dcgm_fb_used:{max5m,avg5m}
- nerc:dcgm_power_usage:{max5m,avg5m}

max5m reliably catches bursts, avg5m keeps the integral correct. The custom allowlist on the hub picks up the new series via ^nerc:dcgm_.*

Scoped to the nerc-ocp-test overlay as a trial; promote to prod/edu by moving the PrometheusRule into nvidia-gpu-operator/base.

### DIFF
--- a/acm-observability/base/core/configmaps/observability-metrics-custom-allowlist/configmap.yaml
+++ b/acm-observability/base/core/configmaps/observability-metrics-custom-allowlist/configmap.yaml
@@ -47,6 +47,10 @@ data:
       - __name__=~"^log_.*"
       - __name__=~"^gpu_operator_.*"
       - __name__=~"^DCGM_FI_.*"
+
+      # Locally pre-aggregated DCGM gauges (max/avg over 5m recording rules),
+      # so the 300s collection interval does not miss short GPU bursts:
+      - __name__=~"^nerc:dcgm_.*"
       - __name__=~"^argocd_.*"
 
       # Consolidate kube*:
--- a/nvidia-gpu-operator/overlays/nerc-ocp-test/kustomization.yaml
+++ b/nvidia-gpu-operator/overlays/nerc-ocp-test/kustomization.yaml
@@ -5,5 +5,6 @@ resources:
   - ../../base
   - configmaps/mig-manager-config.yaml
   - drivers
+  - prometheusrules/dcgm-recording-rules.yaml
 patches:
   - path: clusterpolicy/clusterpolicy_patch.yaml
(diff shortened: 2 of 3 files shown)


schwesig:
Note for reviewers: this PrometheusRule only gets picked up by the platform Prometheus if the nvidia-gpu-operator namespace carries the label `openshift.io/cluster-monitoring: "true"`.
That label is not set in the namespace manifest in this repo, but it must already be present on the cluster.
The DCGM exporter ServiceMonitor in the same namespace is already being scraped by platform Prometheus, otherwise we would not see any DCGM_FI_* metrics in ACM observability at all.

So from my perspective this should just work, but happy to add the label declaratively to cluster-scope/base/core/namespaces/nvidia-gpu-operator/namespace.yaml in this PR if we would rather have it explicit.

computate (approved):

schwesig:
FYI:
The existing gauge metrics stay exactly as they are: still one instantaneous sample every 5 minutes, not averaged. 

What the PR adds are separate, additional series next to them:
nerc:...:avg5m (average over the trailing 5 minutes) and nerc:...:max5m (peak over the trailing 5 minutes).

Those are computed locally on the cluster at full ~30s resolution, before the 5-minute collection happens.
So the one sample that reaches Grafana carries the whole window instead of a random instant.

In Grafana you then pick per panel what you want:
- raw snapshot (as today), 5-min average, or 5-min peak.
- Rule of thumb:
  - "did my job run at all?" → max5m,
  - "how much did it actually use?" → avg5m.

Resolution is still one datapoint per 5 minutes for all of them!!!
**What changes**:
- max5m makes sure short bursts can't fall through the cracks anymore (any activity in the window shows up at full height), and
- avg5m tells you how much of the window was actually used.

gavinsan33 (approved):

IsaiahStapleton (approved):

schwesig:
merged, working now
image

PR #969, opened 2026-07-23
feat(ai-prism-dashboard): v2 with workload row and corrected GPU coun…

schwesig:
Why:
The v1 dashboard showed 40 allocated GPUs while the Barcelona cluster only has 20. Finished (Succeeded) StatefulSet pods still report their GPU requests, and the sparse Barcelona metrics (~1 sample per hour) made the Running-join panels undercount. It was also not visible whether allocated GPUs actually compute. v1 stays untouched, v2 is added next to it in the same folder.

Key changes:
- new file ai-prism-project-dashboard-v2.yaml (own CR, uid ai-prism-project-v2, folder AI-PRISM), added to kustomization
- new top row 'Workload - What is actually running': table of GPU pods (phase, start time), CPU cores in use (3h window, works as idle detector), GPUs held by Running pods, cluster GPU capacity, and a no-join control stat incl. finished pods
- text panel explaining the 20 / 40 / 60 GPU readings (Running, incl. finished pods, restart overlap)
- widen all Running-join windows from 10m to 2h so the sparse Barcelona phase samples are bridged (fixes ~30% undercount in the GPU/CPU hours allocated panels)
- rename row 'AI-PRISM - Per Project (Namespace)' to 'Allocated vs used'

Triggered by: 40 vs 20 GPU allocation question from the team
Connected to: nerc-project/operations#1562 nerc-project/operations#1566

### DIFF
--- a/grafana/overlays/nerc-ocp-obs/grafanadashboards/ai-prism-project-dashboard-v2.yaml
+++ b/grafana/overlays/nerc-ocp-obs/grafanadashboards/ai-prism-project-dashboard-v2.yaml
@@ -1,0 +1,1057 @@
+apiVersion: grafana.integreatly.org/v1beta1
+kind: GrafanaDashboard
+metadata:
+  name: ai-prism-project-dashboard-v2
+  namespace: grafana
+  labels:
+    app: grafana
+spec:
+  instanceSelector:
+    matchLabels:
+      dashboards: grafana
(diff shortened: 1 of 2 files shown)


EldritchJS (approved):
lgtm

memalhot (approved):

computate (approved):

PR #971, opened 2026-07-27
fix(ai-prism-dashboard): correct barcelona sample-rate wording in v2

schwesig:
The two text panels claimed barcelona sends only ~1 sample per hour.
Normal push interval is 300s (ACM metrics-collector default); the sparse data seen earlier came from intermittent transfer gaps of several hours.
The 2h lookback stays, only the explanation is corrected.

---
### more details:

This only changes the wording in the two explanation text panels, no queries or panels touched.

The panels claimed barcelona sends only ~1 sample per hour. Turned out that's wrong as a general statement, I re-checked and barcelona pushes every 300s like the other clusters (ACM metrics-collector default).

The sparse data I saw earlier came from intermittent transfer gaps of several hours (e.g. July 20 evening to July 21 noon), and I just happened to measure during one of those windows.

The 2h lookback in the join queries stays as it is. In normal operation it just picks the newest of ~24 samples (no effect), and during a gap it keeps the panels on the last known state instead of showing 0/no data.

So only the explanation gets corrected here, the mechanism was and still is right.


EldritchJS (approved):
lgtm thanks

computate (approved):

PR #976, opened 2026-08-17
Add GPU reserved-vs-computed dashboard v6

schwesig:
Same three views as v5 plus history support:
stat tiles ask range questions and both time series panels cap their resolution, so the dashboard keeps working for time ranges up to ~12 months back (hub downsampling only serves steps of 25min and coarser).

Adds a "Data aging" doc tile explaining the retention stages.

Data aging

- Metrics live in 3 stages in the hub: raw 300s samples (90 days), 5min downsampled (360 days), 1h downsampled (forever).
- Dashboard is build so all panels still work up to roughly 12 months back, panels ask range queries with a coarse enough step so Thanos serves the downsampled data automatic.
- Rule of thumb: Thanos only serves a downsampling level if step >= 5x the level. 5min data needs step >= 25min, 1h data needs step >= 5h.
- For ranges older than 90 days: numbers are slightly smoothed (5min instead of raw), differences are in the low percent range.
- For ranges older than ~12 months: only 1h data exists, too coarse for these formulas, so panels will be empty there.
- If a panel shows no data for an old range, check the range is under 12 months and that the step isn't forced too fine (zooming very far into old data can push the step under the threshold again).
- The 90/360 day values are not fix, they come from the ACM retentionConfig in GitOps: [multiclusterobservability.yaml (retentionConfig)](https://github.com/OCP-on-NERC/nerc-ocp-config/blob/main/acm-observability/base/observability.open-cluster-management.io/multiclusterobservabilities/observability/multiclusterobservability.yaml#L45-L51). If someone changes retention there, this tile does not update automaticaly, so check there when in doubt.
- Source for the step >= 5x rule: [Thanos docs, Auto downsampling](https://thanos.io/tip/components/query.md/#:~:text=the%20replica%20labels.-,Auto%20downsampling,1h%20%2D%20Use%20max%201h%20downsampling.,-Partial%20Response%20Strategy) (max_source_resolution defaults to step / 5, and 0 means raw only, which is also why instant queries never get downsampled data).

> Side note worth remembering:
> maxDataPoints feels like a display setting,
> but on the Thanos-hub it is realy a data availability knob for old time ranges.

Source:
- for  /5 rule
  - https://github.com/thanos-io/thanos/blob/8908023d563bd5d71ee69b93b6afa857765cccf4/pkg/api/query/v1.go#L826

### DIFF
--- a/grafana/overlays/nerc-ocp-obs/grafanadashboards/gpu_ReservedVsAttachedVsComputed/gpu-reserved-vs-computed-v6.yaml
+++ b/grafana/overlays/nerc-ocp-obs/grafanadashboards/gpu_ReservedVsAttachedVsComputed/gpu-reserved-vs-computed-v6.yaml
@@ -1,0 +1,667 @@
+apiVersion: grafana.integreatly.org/v1beta1
+kind: GrafanaDashboard
+metadata:
+  name: gpu-reserved-vs-computed-v6
+  namespace: grafana
+  labels:
+    app: grafana
+spec:
+  instanceSelector:
+    matchLabels:
+      dashboards: grafana
+  folder: GPU Accounting
+  json: |-
+    {
+      "panels": [
+        {
+          "gridPos": {
+            "h": 2,
+            "w": 24,
+            "x": 0,
+            "y": 0
+          },
+          "options": {
+            "content": "# NVIDIA GPUs only (Filter nvidia.com and DCGM metric)\n\nreservation vs. real useage, per cluster and namespace (see variable and panel info icons for details)",
+            "mode": "markdown"
+          },
+          "title": "",
+          "transparent": true,
+          "type": "text"
+        },
+        {
+          "datasource": {
+            "type": "prometheus",
+            "uid": "a0fa9d88-8932-41f7-af80-6f678d4fb1e7"
+          },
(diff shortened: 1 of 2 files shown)


schwesig:
see demo here

https://grafana.apps.obs.nerc.mghpcc.org/d/gpu-trio-v6/nvidia-gpu3a-reserved-vs-attached-vs-computed-v6?from=2026-02-01T00:00:00.000Z&to=2026-02-28T23:59:59.000Z&timezone=utc&var-cluster=$__all&var-namespace=multi-modal-semantic-routing-for-vllm-d266df&var-idle_window=4h&var-util_threshold=1&refresh=1h

image

schwesig:
For review context: the diff between v5 and v6 looks big (new file), but on the query side it is realy small. All PromQL expressions are 1:1 identical to v5 (checked programmatic, every `expr` string unchanged). What changed is only HOW the queries get executed:

**The 3 stat tiles** (reserved / attached / computed): instant query becomes range query, with few data points:

```diff
   "expr": "sum(sum_over_time(kube_pod_resource_request{...}[$__range])) * 300 / 3600",
-  "instant": true
+  "range": true
+  (panel) "maxDataPoints": 20
```

**The 2 time series panels** (cumulative curves, idle detector): resolution capped:

```diff
   "expr": unchanged
+  (panel) "maxDataPoints": 800
```

**Why:** the hub only serves downsampled data (5min level, kept 360 days) when the query step is at least 5x the level, so step >= 25min. Instant queries have no step, they always get raw only (kept 90 days). That is why v5 showed "No data" for time ranges older than ~90 days: tiles were instants, and the curves in pixel resolution had steps around 20min, just under the threshold. v6 fixes both, so the dashboard works up to ~12 months back now. The new "Data aging" doc tile at the bottom explains this for users, with links to the retentionConfig in this repo and the Thanos docs as source.

Plus one new panel (the doc tile), no other changes.

schwesig:
Small follow up on the "HOW the queries get executed" part, because the mechanic behind it is worth knowing:

Thanos does not look at the PromQL at all when it decides which storage level to serve. It only looks at the request parameters. The chain:

```
panel setting             HTTP request to Thanos                 Thanos decision
instant: true         ->  /query       (no step)             ->  step/5 = nothing -> raw only (90 days)
range + maxDataPoints ->  /query_range (step = range/points) ->  limit = step/5 -> pick the level
```

| panel | before (v5) | after (v6) |
|---|---|---|
| stat tiles | instant, no step -> raw only | range, 28d/20 = 33.6h step -> limit 6.7h -> even the 1h level qualifies ✓ |
| curves / idle detector | ~1500 points -> step ~20min -> limit 4min < 5min ✗ | 800 points -> step ~50min -> limit 10min >= 5min ✓ |

So the formulas compute the same thing in both versions, they just actually get data delivered now.


> Side note worth remembering:
> maxDataPoints feels like a display setting,
> but on the hub it is realy a data availability knob for old time ranges.


Source: Thanos auto downsampling, `max_source_resolution` defaults to step/5: https://thanos.io/tip/components/query.md/

computate (approved):

EldritchJS (approved):
lgtm

PR #977, opened 2026-08-19
Add GPU reserved-vs-computed dashboard v7

schwesig:
## What

v7 of the GPU reserved vs attached vs computed dashboard. Only change vs v6, the three stat tiles do not calculate on there own anymore, they sum up exactly the increments of the cumulative curves below (stat reduce: Total). This fixes the step alignment mismatch, where the tile showed 441.6 GPU-h but the curve ended at 691 for the same June range; now the tile shows 690.9, matching the curve.

## Note

#976 (v6) is merged, so this PR shows only the v7 commit now. Next in the chain: #978 (v8) and #979 (v9) build on top of this one, so best merge this one first.

A demo of v7 is deployed in the obs Grafana, folder "(deleted after restart) > 2026-08-05_gpuReservedVsAttachedVsComputed", dashboard [/d/gpu-trio-v7](https://grafana.apps.obs.nerc.mghpcc.org/d/gpu-trio-v7).


EldritchJS (approved):
lgtm

computate (approved):

PR #978, opened 2026-08-19
Add GPU reserved-vs-computed dashboard v8

schwesig:
## What

v8 of the GPU reserved vs attached vs computed dashboard. Adds a dynamic "DCGM coverage by cluster" warning table at the bottom, one row per ACM managed cluster (list comes dynamic from acm_managed_cluster_labels, no static list), telling why the attached/computed panels can be empty or too low.

States:
- SILENT purple (-1), cluster sends nothing to the hub right now
- blue (-2), cluster has no GPU nodes, nothing to messure
- green (0), has GPU nodes and all of them send DCGM
- orange (1-99), percent of GPU nodes without DCGM, numbers are a lower bound
- red (100), no DCGM at all, attached/computed stay empty

One server side score query, no client side joins. Plus a legend text tile below that explains all five states. This replaces the old static "barcelona has no DCGM" note with something that stays correct on its own.

## Note

Stacked on #977 (v7): this PR is based on the feat/gpu-dashboard-v7 branch, so it shows only the v8 change. Merge #977 first; after that this PR gets retargeted to main.

A demo of v8 is deployed in the obs Grafana, folder "(deleted after restart) > 2026-08-05_gpuReservedVsAttachedVsComputed", dashboard [/d/gpu-trio-v8](https://grafana.apps.obs.nerc.mghpcc.org/d/gpu-trio-v8).


EldritchJS (approved):
+1

computate (approved):

PR #979, opened 2026-08-19
Add GPU reserved-vs-computed dashboard v9

schwesig:
## What

v9 of the GPU reserved vs attached vs computed dashboard. Builds on the v8 coverage table and adds the time dimension.

- The coverage table gets a second column. "now" is the state at the end of the selected range (the fix it view, and the only place where SILENT shows up reliable). "window avg" is the percent of GPU node time inside the range without DCGM (the trust the numbers view). Every window avg above 0 means attached, computed and the idle detector are a lower bound for that cluster.
- A small stat tile on top, right under the header, shows the window avg for the selected clusters only, so you see immediatly how much to trust the numbers below (lower is better).
- The avg only counts time where the cluster was reporting to the hub, partial silence shows only in the now column. Both run as instant queries, so the usual 90d raw retention limit applies.

Why: the instant only view could show 33% for a cluster where only 15% of the GPU node time in the window was actually blind (or 0% right after an exporter came back). The average answers the real question of this dashboard, how reliable are the tiles for the selected range.

## Note

Stacked on #978 (v8): this PR is based on the feat/gpu-dashboard-v8 branch, so it shows only the v9 change. Merge #977 and #978 first; after that this PR gets retargeted.

A demo of v9 is deployed in the obs Grafana, folder "(deleted after restart) > 2026-08-05_gpuReservedVsAttachedVsComputed", dashboard [/d/gpu-trio-v9](https://grafana.apps.obs.nerc.mghpcc.org/d/gpu-trio-v9).


EldritchJS (approved):
lgtm

computate (approved):

